Issue #16, opened 2015-09-16
Add the whole CI thing


Drup:
A decent way to test is to compile `mirage` and `mirage-skeleton`

@samoht, could you write the runes ? :]

PR #17, opened 2015-09-16
Minor admin stuff

samoht:
Don't merge yet, I'll add the travis scripts as well


Drup:
You need to pin fmt too.

samoht:
Hum this doesn't compile with 4.01. Not sure we are ready to drop 4.01 support yet.

Drup:
probably some "module M = M'" in some .mli, let me see.

samoht:
yes this is indeed due to the module aliases. I'm wondering if it makes sense to have only one public .mli instead of spread the documentation over 3 and use aliases.

(the opam file also miss some availability constraint)

Drup:
4, not 7, the others are internal. And I do prefer to have the interface associated with its module instead of bundled together in one big .mli that is a pain to edit as it goes bigger and more messy. The documentation will have hyperlinks anyway, so I don't see the issue.

samoht:
I have no strong opinion on this, although it is usually easier to read and better organised (for the library user) when everything is in one place.

Drup:
Can you rebase and put the appropriate constraints ?

Drup:
Also, to solve #16, I would really like to automatically download and build mirage-skeleton.

samoht:
I've rebased. It's a bit hard currently to clone and test mirage-skeleton because it needs a new release of mirage. I'll add some runes anyway that could be cleaned-up later.

Drup:
I missed a bit for compat. It's fixed. you can rebase.

samoht:
rebased

Drup:
It's green \o/

Can't we use REVDEPS instead of the manual opam launch ?

samoht:
Good points for REVDEPS, updated.

Drup:
Hum, I realized this actually doesn't build mirage-skeleton. The opam package apparently does nothing. why is that ?

samoht:
Hum... I think that's because you cannot call Opam inside Opam. Need to double check.

samoht:
All green! (for real this time)

Drup:
Yay! And it's testing xen too, which is really useful. Thanks!

PR #19, opened 2015-09-17
Restore compat with 4.01

Drup:
Minimal changes to make it work. those `module type of struct include` annoys me to no end.

### DIFF
--- a/src/functoria.mli
+++ b/src/functoria.mli
@@ -18,7 +18,7 @@
 (** Configuration library *)
 
 (** Core functoria DSL *)
-module Dsl = Functoria_dsl
+module Dsl : module type of struct include Functoria_dsl end
 
 (** Various generic devices. *)
 module Devices : sig
@@ -55,7 +55,7 @@ module Devices : sig
 end
 
 (** Various helpful functions. *)
-module Misc = Functoria_misc
+module Misc : module type of struct include Functoria_misc end
 
 (** A specialized DSL build for specific purposes. *)
 module type SPECIALIZED = sig
--- a/src/functoria_dsl.mli
+++ b/src/functoria_dsl.mli
@@ -54,7 +54,7 @@ val hidden : _ impl -> any_impl
 (** {2 Keys} *)
 
 (** Key creation and manipulation. *)
-module Key = Functoria_key
+module Key : module type of struct include Functoria_key end
 
 val if_impl : bool Key.value -> 'a impl -> 'a impl -> 'a impl
 (** [if_impl v impl1 impl2] is [impl1] if [v] is resolved to true and [impl2] otherwise. *)
--- a/src/functoria_key.ml
+++ b/src/functoria_key.ml
@@ -138,7 +138,7 @@ module Set = struct
     type t = elt
     let compare (Any k1) (Any k2) = compare k1.name k2.name
   end
-  include (Set.Make (M) : Set.S with type elt := elt)
+  include (Set_Make (M) : SET with type elt := elt)
 
   let add k set =
     if mem k set then
--- a/src/functoria_key.mli
+++ b/src/functoria_key.mli
@@ -65,7 +65,7 @@ module Doc : sig
 
 end
 
-module Set : Set.S
+module Set : Functoria_misc.SET
 (** A Set of keys. *)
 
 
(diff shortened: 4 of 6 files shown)


samoht:
why not simply `module type of X`?

avsm:
I think this is a reasonable time to go 4.02-only for Mirage..

mor1:
@avsm Out of interest, does anyone try to maintain stats on use (or
perhaps, "reach") of the different OCaml compiler versions?

On 17 September 2015 at 17:00, Anil Madhavapeddy <notifications@github.com>
wrote:

> I think this is a reasonable time to go 4.02-only for Mirage..
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mirage/functoria/pull/19#issuecomment-141132515>.
>



-- 
Richard Mortier
mort@cantab.net

samoht:
I think we should keep compat as long as it not too much a pain. In that case, the changes are minimal, although not very pretty so that's fine.

Also, I think it's good to keep supporting the last 2 major versions of OCaml, to let people evolve their code-base.

avsm:
Not to my knowledge. The only source of this is the OPAM compiler-version constraints.

mor1:
Ack. I guess that analysis would be a start, to see how fast package
maintainers are switching.
Not even download stats from places where compiler is installed from?

On 17 September 2015 at 17:10, Anil Madhavapeddy <notifications@github.com>
wrote:

> Not to my knowledge. The only source of this is the OPAM compiler-version
> constraints.
>
> > On 17 Sep 2015, at 17:08, Richard Mortier <notifications@github.com>
> wrote:
> >
> > @avsm Out of interest, does anyone try to maintain stats on use (or
> > perhaps, "reach") of the different OCaml compiler versions?
> >
> > On 17 September 2015 at 17:00, Anil Madhavapeddy <
> notifications@github.com>
> > wrote:
> >
> > > I think this is a reasonable time to go 4.02-only for Mirage..
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <https://github.com/mirage/functoria/pull/19#issuecomment-141132515>.
> > >
> >
> >
> >
> > --
> > Richard Mortier
> > mort@cantab.net
> > —
> > Reply to this email directly or view it on GitHub <
> https://github.com/mirage/functoria/pull/19#issuecomment-141134844>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mirage/functoria/pull/19#issuecomment-141135201>.
>



-- 
Richard Mortier
mort@cantab.net

avsm:
Fine by me.

dbuenzli:
Le jeudi, 17 septembre 2015 à 17:17, Richard Mortier a écrit :
> Not even download stats from places where compiler is installed from?

I don't think we give a damn about download stats, it's a flawed metric anyways. What Thomas proposes is the right way of proceeding. I.e. say you support at least X version before the latest release. Given that OCaml major releases are semi-predictable it gives you, from a software maintenance perspective, a timeframe to plan your migration and whip programmers.

(This however doesn't the consider the question of package managers lingering with old ocaml versions which you may want to put in the equation, depends a bit on the kind of apps you develop)

Best,

Daniel

mor1:
Flawed metric? Depends what's being inferred.

If there was some way to track the rate of installation of new compiler releases, as eg Google do with Android versions, then you might get some idea of the rate of uptake of a new release and what proportion of the community you're losing by forcing an upgrade.

Ditto looking at OPAM package repo and seeing how many (or better, rate at which) packages are mandating moving to newer releases.

Drup:
@samoht 

if you do `module M = module type of include X end` then for all type `t` in `X`, `M.t = X.t`. This is not the case without the `include`. 

Yes, that's a feature of the OCaml module system...

dbuenzli:
Le jeudi, 17 septembre 2015 à 17:48, Richard Mortier a écrit :
> Flawed metric? Depends what's being inferred.

The data itself is flawed. Garbage in, garbage out.

> Ditto looking at OPAM package repo and seeing how many (or better, rate at which) packages are mandating moving to newer releases.
That only tells you something about package developers (which are more keen on using the latest shiny features), not package *users*. If it were only me I would rewrite cmdliner using gadts, but I have a very important user that needs 3.12.0.

Using these numbers completely misses the point of being able to derive timeframes for maintenance which is what you want from a user point of view. You don't want support to be kept or dropped based on whathever broken metric one can invent.

Best,

Daniel

PR #24, opened 2015-10-03
Various improvement to keys

Drup:
This adds:

Env setting with latest cmdliner. I choose to have something more simple, documentation wise, by just using the default (which redirects to the key). I personally thing it's far enough.
Otherwise, not much to say, it works perfectly well, including at runtime.

flag and proxy keys: Wow, that was very painful. The internal is much more powerful than what is exposed, but what is exposed is the only part that doesn't have any weird quiks.

The priority is well behaved, if `--deploy` set, among other things, `ip` and you do `--deploy --ip FOO`, it's going to work as expected. I do not garantee *anything* if two proxies step on each other toes (not
even a warning, at the moment ...).

One weirdness, which is somehow expected but still slightly annoying, is that if I have `--xen` setting `-t xen`. Then `Bootvar_gen.xen () = true` only if `--xen` was passed. Not if `-t xen` was passed.

It makes sense, because of the deploy example I gave previously.

What is forbidden by the API because too quirky:
- non-flag proxies, because I have no idea what semantic to give them, when no value is given.
- runtime proxies, because it's not going to be possible with printf-CSP


samoht:
CI should be fixed by https://github.com/ocaml/ocaml-travisci-skeleton/pull/71

PR #25, opened 2015-10-07
Add a --color command-line option

### LINKED ISSUE #18: force colors?
Seems that there are no color anymore when the mirage tool is called inside a Makefile. I guess the tty auto-dectection does not handle very well that case. 

Maybe add an option to `mirage configure` to force the display or color (using the usual tri-state logics?)

samoht:
Fix #18

Inital code from @dbuenzli

Not sure if that the simplest way to do this


Drup:
1. Look at your warnings. :]
2. Yes you can do simpler: hook it up directly inside `with_config`. You need to execute the init color before loading the config file for this to work. Don't forget to add it to the help subcommand too.

src/functoria_tool.ml
```diff
@@ -78,7 +78,14 @@ module Make (Config : Functoria_sigs.CONFIG) = struct
     in
     Arg.(value & opt (some string) None & doc)
 
-  let () =
+  let color =
+    let enum = ["auto", None; "always", Some `Ansi_tty; "never", Some `None] in
+    let color = Arg.enum enum in
+    let alts = Arg.doc_alts_enum enum in
+    let doc = Fmt.strf "Colorize the output. $(docv) must be %s." alts in
+    Arg.(value & opt color None & info ["color"] ~doc ~docv:"WHEN")
```
Drup: Please also put the right documentation section (see the various options above).

samoht:
I've updated my patches, should be simpler now.

_tags
```diff
@@ -31,6 +31,7 @@ true: annot, bin_annot
 # OASIS_STOP
 
 true: warn(+A-4-6-7-9-40-42-44-48)
+true: warn_error(+1..49)
```
Drup: Doesn't this mean it will trigger for 4,6, ... ? I'm not familiar with warn error ...

Drup:
:+1:

_tags
samoht: @damiendoligez said so :p https://github.com/mirage/irmin/commit/b0b6bbe6d7a18b193171afc7aad235059f134b4b

Drup: So the semantic of warn_error is "if it warns, it errors" and not "enable those warning and turns them into errors" ? This is a bit strange but ok.

samoht:
My plan now is to improve the generated doc. I'm not very fond of the module renaming and inclusion game that you are playing everywhere in the codebase, I might change that as well.

_tags
damiendoligez: If you want the latter, you write `-w @1..49`.

Drup: It's actually the good one here. Thanks for the precision!

samoht:
@Drup do you mind if I merge that PR? more changes are coming in soon, which will be much more invasive than this one :-)

Drup:
Go on. You started adding stuff unrelated to the initial issue, so I let it sleep.

PR #26, opened 2015-10-16
Api changes

samoht:
The engine didn't change it all, just trying to follow Cmdliner API more closely. Also tried to document most of the API in a consistent way. Not totally done yet, but I think it start to make sense as an independent project.


Drup:
Yay, someone else than me got the courage to rewrite `Functoria_misc`! :D 

What I like:

- The `Functoria` "misc" submodules. It's what I wanted to do since the beginning, but got lazy about it, thanks. :D
- The split between lib and app, it should have been that way since the beginning, it's nicer.
- Most of the new names (context and alias, in particular)

What I don't like: 

- `Key.v` is begging for shadowing issues. Why not use the same convention than Impl (it was `hidden`, you renamed it to `abstract`, which is good to).
- I'm not convinced it's a good idea to expose `explode`.
- I'm not sold on the `Alias` interface. The interface was very constrained to avoid weird things in the engine, do you respect those ? (Arguably, the constraints are not documented ...) 
- Why is `converter` not a record anymore ? Unnamed tuples are bad.
- I *really* prefer `Functoria.Info.libraries/packages` to be a set. You seem to have removed it in order to not have to expose a String Set module, and I find that a rather poor reason to provide a strictly less efficient and less expressive API.
- You removed some module aliases that are helpful for the specialized DSL. In particular, in the updated mirage, you needed to add [this](https://github.com/samoht/mirage/blob/functoria/lib/mirage.mli#L27) and [this](https://github.com/samoht/mirage/blob/functoria/lib/mirage_key.mli#L17). Those signatures *should* be defined in functoria, because the user is going to get them wrong (proof: pretty sure the first one is wrong).

I don't like the names in cmdliner but you choose the consistency. Why not.

samoht:
> Key.v is begging for shadowing issues. Why not use the same convention than Impl (it was hidden, you renamed it to abstract, which is good to).

In a previous version I used `Key.dep`. Do you like it better?

> I'm not convinced it's a good idea to expose explode.

ok I'll remove it. If you see other things to remove, please shout.

> I'm not sold on the Alias interface. The interface was very constrained to avoid weird things in the engine, do you respect those ? (Arguably, the constraints are not documented ...)

What are the constraints? I'm not sure about that as well, but the I really didn't find a good way to explain the previous `Setter` API.

> You removed some module aliases that are helpful for the specialized DSL. In particular, in the updated mirage, you needed to add this and this. Those signatures should be defined in functoria, because the user is going to get them wrong (proof: pretty sure the first one is wrong).

ok, I'll re-add them.

Thanks for the quick review!

Drup:
> In a previous version I used Key.dep. Do you like it better?

Meh, It doesn't have much to do with dependencies.

> What are the constraints? I'm not sure about that as well, but the I really didn't find a good way to explain the previous Setter API.

Let me look at things more closely

> ok I'll remove it. If you see other things to remove, please shout.

I didn't see any. The serializing functions in `Functoria_key` are okish to expose, since they won't break anything.

samoht:
`Key.dep` create `t` which is documented as a data-dependency. `Key.abstract` would work as well, but that's yet again a new concept to explain :-) `Key.create` maybe.

Drup:
`Key.create` is already taken. 
My issue with `Key.dep` is that the parallel with `Key.deps` ... doesn't exist.

dbuenzli:
> Key.create is already taken. 
> My issue with Key.dep is that the parallel with Key.deps ... doesn't exist.

Something like `Key.value_{to,as}_dep` or even `Key.to_dep` ?

dbuenzli:
In fact the description of this `t` is a little bit strange. Isn't that simply a `key` whose parameter is hidden ? In this case `Key.hide_type` even though a bit artificial may do it.

samoht:
ok thanks, API is updated. Still need to improve `Alias`.

Also, the `-f path/to/config` option in mirage doesn't seem to work anymore, I should have done something wrong with the refactoring. I'll try to fix that later...

samoht:
> Why is converter not a record anymore ? Unnamed tuples are bad.
> I really prefer Functoria.Info.libraries/packages to be a set. You seem to have removed it in order to not have to expose a String Set module, and I find that a rather poor reason to provide a strictly less efficient and less expressive API.

no strong opinions on those, although I really don't like a specialised library (as Functoria) to expose a `StringSet` module (which should live somewhere else, maybe in `Astring`?)

for the record, I kind of agree. I've tried to follow [Cmdliner] API which defines a converter to be a pair. I'm fine to switch back to a record.

Drup:
> In fact the description of this t is a little bit strange. Isn't that simply a key whose parameter is hidden ? In this case Key.hide_type even though a bit artificial may do it.

Well, before @samoht's patches, it was called `hidden` (suggestion by @mor1) so that you write `Key.[hidden foo ; hidden bar]`. Same things for the equivalent `impl`. Now the equivalent `impl` thing is called `abstract`, and I really don't see a reason not to have the same name.

dbuenzli:
> it was called hidden (suggestion by @mor1)

I think `hidden` or `hide` is a good name. 

> which should live somewhere else, maybe in Astring?

String sets do exist there, but then you may not want to introduce the dependency only for this.

> I've tried to follow [Cmdliner] API which defines a converter to be a pair. I'm fine to switch back to a record.

Since it's strictly speaking not the same type as `Cmdliner` I don't see any advantage of following what is done there. Mark Shinwell would scream anyway and say that this should be kept abstract, which I kind of agree.

Drup:
> Mark Shinwell would scream anyway and say that this should be kept abstract, which I kind of agree.

I agree too (it was abstract before, with an explicit constructor)

Drup:
I'm looking into keys more carefully now:

- I think we should move the default value from `Arg.t` to `Arg.kind`, in the `Opt` constructor. In particular it opens the way for easy implementation of the mandatory arguments.
- `Arg.serialize` ~default argument is not the default, it's the value. I'm not convinced it should be optional (in all circumstances inside functoria, it's incorrect not to handle it).
- The `Arg` module is now a bit big and not particularly well split into subparts ...
- `Arg.info`'s documentation should mention the default section.
- `Key.value`'s documentation is ... wrong. It doesn't have much to do with parsing. I'm not sure what the goal was. I think the paragraph about data-dependencies should be there, instead of `t`. Globally, I'm really not convinced by your take on the documentation about the relation between the types `value`/`key`/`t`.
- The section `Alias` doesn't appear in the generated documentation. It's considered as header for the `Alias` module instead.
- `Context` should probably be a submodule.
- In the runtime, the `name` field in `Key.t` is unused.
- In the runtime, documentation is not up to date.

I finally made sens of your version of `Alias` ... I don't think having the `info` inside is easy to understand, but maybe that's because I had the other interface in mind. The previous version disallowed `Alias.opt`, since having a default value makes things weird (the alias is triggered with nothing on the command line, just with the default value).

samoht:
> I think we should move the default value from Arg.t to Arg.kind, in the Opt constructor. In particular it opens the way for easy implementation of the mandatory arguments.

seems ok.

> Arg.serialize ~default argument is not the default, it's the value. I'm not convinced it should be optional (in all circumstances inside functoria, it's incorrect not to handle it).

true, I'll fix that.

> The Arg module is now a bit big and not particularly well split into subparts ...
> Arg.info's documentation should mention the default section.

I really want [Arg] to be a direct extension of [Cmdliner.Arg] (with more complex converters) and a `stage` argument to the creators. Users already know that API, let's not force them to learn a new one for doing "almost" the same thing.

> Key.value's documentation is ... wrong. It doesn't have much to do with parsing. I'm not sure what the goal was. I think the paragraph about data-dependencies should be there, instead of t. Globally, I'm really not convinced by your take on the documentation about the relation between the types value/key/t.

Agreed this is still confusing. But if you have a better wording, patches and/or improvements are welcome!

> The section Alias doesn't appear in the generated documentation. It's considered as header for the Alias module instead.
> Context should probably be a submodule.
> In the runtime, the name field in Key.t is unused.
> In the runtime, documentation is not up to date.

thanks, I'll fix these (not sure about the `Context` submodule, I've tried it and didn't like the result much)

Drup:
> I really want [Arg] to be a direct extension of Cmdliner.Arg and a stage argument to the creators. Users already know that API, let's not force them to learn a new one for doing "almost" the same thing.

I'm actually talking about the implementation here, not the interface. Everything is pretty much pilled into a big blob, in no real order. At least order them in a coherent fashion and put sections.

samoht:
I tried to address most of you concerns.

> Arg.info's documentation should mention the default section.

Not sure what to do here: `Arg.info` does not contain the `default` value, only an `'a t` does.

Drup:
I meant the default _section_, `"UNIKERNEL PARAMETERS"`.

samoht:
@Drup I think I don't understand how the `describe` thing work.

Drup:
What makes you think `describe` is broken ? It works fine for me. 

To make the behavior clearer: By default it will try to call `xdot` and feed it the dot in stdin. It can use something else than `xdot` with `--dot-command` and with `-o` it will dump the dot in a file. If `--dot` is not given, and since `pp` is a noop at the moment, it prints nothing. :]

Drup:
Grmbl, by copying the interface in cmdliner, we lost the labels on the `Arg` functions. ಠ_ಠ

dbuenzli:
> Grmbl, by copying the interface in cmdliner, we lost the labels on the Arg functions. ಠ_ಠ

By all means, do not copy what is not good.

> I don't like the names in cmdliner but you choose the consistency. Why not.

Just curious, which ones are these ?

Drup:
`opt` and `some` are quite bad names, but that's not a very big concern. My biggest issue is that the `Arg` module mixes different concerns (the documentation, the converter and the form of the command line). In my version of the API, I had a `Doc` and a `Desc` (converters) submodules to isolate those concerns.

I also think it works better with functoria, since mirage (or any specialized dsl) will want to extend the module containing the converters for some special purposed ones, but the rest stays the same

My impression was also that, contrary to what @samoht seems to think, people don't actually remember this API and have to rediscover it every time they need to compose a new command line (it doesn't happen every day ...). The fact that `Arg` has many functions with different purposes doesn't help there. 
(But we should clearly have "API surveys" ;)

(however @samoht seems to care a lot about that one, I disagree enough to grumble, but consistency is a decent enough argument :p)

dbuenzli:
> My biggest issue is that the Arg module mixes different concerns (the documentation, the converter and the form of the command line). 

What I know is that these `info` are terrible names. Regarding the mix of concerns it also depend a bit on usability --- avoiding too many names and make it easy to use through the `M.()` notation. But I wouldn't claim that what has been done there is the best possible.

> My impression was also that, contrary to what @samoht seems to think, people don't actually remember this API and have to rediscover it every time they need to compose a new command line (it doesn't happen every day ...). 

I think this is true. A lot of people tell me they often cut and paste cmdliner code which is suspicious and that maybe some simpler interface should be provided on top of it --- but I'd still stand behind the full power of the current interface, especially the for its composable aspects.

> (But we should clearly have "API surveys" ;)

I'd prefer it to be called "API trashing" so that every one out there doesn't dare to take its own shit at hart.

Drup:
I do agree on the composable aspect, even if I did terrible things with it in functoria (where is my monadic bind ?!). I think people in the OCaml ecosystem are even less used to applicatives than they are to monads.

dbuenzli:
> some are quite bad names,

Btw. would you have another name for [this](http://erratique.ch/software/cmdliner/doc/Cmdliner.Arg.html#VALsome) or  [that](http://erratique.ch/software/jsont/doc/Jsont.html#VALsome) ? 

> (where is my monadic bind ?!)

Lack of monadic bind is what allows cmdliner to generate accurate manpages.

Drup:
I would just call it `option`.

dbuenzli:
> I would just call it option.

This a little bit OT and a discussion that should happen on the API trashing mailing list (which we should definitivitely open for some fun and quality improvement) but note that in both cases it's not an option parser/decoder which is why I'm not using `option`: all other `t` combinators do parse data representing the `t` type. In both cases it's a combinator for lifting *absence/presence* into option.

Drup:
Do open the mailing list, and I will answer there. :D

Drup:
@samoht I updated mirage-skeleton. Everything seems to work fine now.

samoht:
great! I'm trying to fix the textual describe and fix your latest comment, ask a final review of the API and then merge...

Drup:
Please add back the sets in the API that you removed.

samoht:
I'm re-adding the key sets, cleaning up the logging verbosity and finishing up the describe command.

samoht:
I've re-added the Set API. Can you clarify how it could help solving the issue with duplicate arguments in Cmdliner? Is it just a matter of failing earlier with a clearer error message?

samoht:
The raw description now prints something vaguely useful:

```shell
~/git/mirage-skeleton/network$ mirage describe
Name       network
Keys       foo-dhcp=false (default), foo-gateways=10.0.0.1 (default), foo-ip=10.0.0.2 (default), foo-net=direct (default), foo-netmask=255.255.255.0 (default), network=tap0 (default),
           target=macosx (default), unix=false, xen=false
Libraries  functoria.runtime, lwt.syntax, mirage-clock-unix, mirage-console.unix, mirage-net-macosx, mirage-types.lwt, mirage.runtime, sexplib, tcpip.arpv4-unix, tcpip.ethif-unix, tcpip.ipv4-unix,
           tcpip.stack-direct, tcpip.tcp, tcpip.udp
Packages   functoria, lwt, mirage-clock-unix, mirage-console, mirage-net-macosx, mirage-unix, sexplib, tcpip
```

Drup:
Clearly, you missed https://github.com/samoht/functoria/pull/2

samoht:
The [online docs](http://mirage.github.io/functoria/) are now updated.

samoht:
@Drup woops...

Drup:
I think you went a bit overboard with the verbosity settings (it prints *nothing* by default).

Otherwise, everything looks good to me. I propose to merge in order to have the big changes in and iron out the last few things in another PR.

samoht:
I'll fix the verbosity settings, but I find the `mirage` tool too verbose currently.

Drup:
Could you go through the various issues on functoria and pick what you want before release/give your opinion ? thanks.

samoht:
I see no blockers for a release. Maybe fix the verbosity issue but I'm fine to have the minimum amount of noise.

Drup:
@samoht I re-did mirage-www with the new key API ... yeah, I definitely want my labels back. The new API is just annoying. Also, having `info` in the Arg module doesn't play nice with local open (you can't write `let info = ... in Key.(create "blob" Arg.(opt string "foo" info))`).

PR #33, opened 2015-10-30
Implement required keys

### LINKED ISSUE #15: Required keys
At the moment, all the keys are optional (and need to have a default value). Apparently, people also want required keys, without a default value.

The good thing is that it should be possible to implement it with barely changing the API, since `default` already has a label.

Drup:
Closes #15

I managed to avoid the `invalid_arg` by stating in the API that required keys were of type `'a option t`, which forces `get` to return an option.
This was slightly un-intuitive but it makes everything clicks into place much more smoothly.

@samoht: Ideas for a good documentation for `Arg.required` ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ script: bash -ex .travis-ci.sh
 env:
   global:
   - PACKAGE=functoria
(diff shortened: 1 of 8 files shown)


samoht:
Yay!

samoht:
`Arg.required` description looks good.

Issue #37, opened 2015-11-01
Error reporting issues


Drup:
Error reporting is wrong when the `config.ml` doesn't compile **and** there is a specific option on the command line. functoria will report that the command line option is not valid, which is quite confusing. 

This issue is tricky to solve because we are doing weird things with cmdliner. I opened https://github.com/dbuenzli/cmdliner/issues/36 to track it there.

Drup:
@dbuenzli This bug is proving very problematic for users and the only way I find to improve the error reporting are really bad hacks. Could you take a look ?

dbuenzli:
Not exactly sure what you are doing but I don't see why you are not able to separate compilation from term evaluation. It seems to me that you should be able to seperate the two and report errors appropriately if the compilation fails, IIRC this actually worked quite well in assemblage.

dbuenzli:
The driver initialisation that takes care of interpreting the assemble.ml files is [here](https://github.com/samoht/assemblage/blob/master/lib-driver/assemblage_driver.ml#L392-L403). The interface is explained [here](https://github.com/samoht/assemblage/blob/master/lib-driver/assemblage_driver.mli#L161-L194). 

But basically the idea is that you should compile during peek and return a term that either [okays or errors](https://github.com/samoht/assemblage/blob/master/lib-driver/assemblage_driver.ml#L402-L403) according to compilation. Now this term *must* be used further down as the first term of any `eval` you may do after, this will ensure that you bail out with an error before further evaluations of the command line are done.

Drup:
> Now this term must be used further down as the first term of any eval you may do after, this will ensure that you bail out with an error before further evaluations of the command line are done.

My point is that you should provide a better API inside cmdliner to deal with things like that. We should discuss this on the ticket on cmdliner.

dbuenzli:
The point is that you can solve the problem now correctly. The second bullet you mention in the cmdliner is exactly what I'm suggesting above. So you can already do this an I'm not sure it's worth to add more to an already complex system that we are pushing to its limits.

PR #39, opened 2015-11-04
Improve errors

Drup:
Should solve https://github.com/mirage/mirage/issues/459 and https://github.com/mirage/functoria/issues/37
The second fix is slightly hackish. I would prefer to do it another way, but until cmdliner improves
support from peek eval errors, I don't see another solution.


talex5:
Thanks - that works for me!

hannesm:
travis is broken due to zarith, https://github.com/ocaml/opam-repository/pull/5081 solves this (if someone merges @samoht ?)

samoht:
zarith PR merged, build restarted

samoht:
All green, merging!

PR #41, opened 2015-11-21
Fix for get_mode and provides a better solution.

Drup:
I was hesitant to expose foreign, but seing @talex5's [bug report](https://github.com/mirage/mirage/issues/470), I think it's better to expose it. It means there is a clear "upgrade" path when you have a foreign but want to add more things to it.

@samoht: Why wasn't Log.warn already there ? I used it for some debuging, so I left it there.


Drup:
@samoht Do you want to review or do I merge ?

samoht:
looks good, thanks!

PR #42, opened 2015-11-30
Fixes for OCaml 4.03

samoht:


Drup:
Hum, it seems like the ascii version is not available on older compiler. Either we silence the warning, or we cppo.

samoht:
eerk

samoht:
trying an other strategy to silence the warnings.

Drup:
Yes, I think it's a better solution.

PR #43, opened 2015-12-04
Functoria_tool refactoring

yallop:
Following @Drup's [suggestion](https://github.com/mirage/mirage/issues/472#issuecomment-161034146) to
> untangle the pile of spagetti that is functoria_tool

here's a start at a refactoring.

The changes are probably uncontroversial:

* Adding a signature (0779422)
* Switching from a functor with side effects and an ignored return value to a function (2ebc289)
* Lifting code that doesn't depend on a functor argument (b63f47c), i.e. changing
  ```ocaml
  module F(X:S) =
  struct
    let f x = e
    ...
  end
  ```

  to

  ```ocaml
  let f x = e
  module F(X:S) =
  struct
    ...
  end
  ```
* Moving side effects out of a functor body (89e1bea), i.e. changing
  ```ocaml
  module F(X:S) =
  struct
    ...
    let () =
      f ()
  end
  let g () = let module M = F(X) in ()
  ```

  to

  ```ocaml
  module F(X:S) =
  struct
    ...
  end
  let g () = let module M = F(X) in M.(f ())
  ```

### DIFF
--- a/app/functoria_tool.ml
+++ b/app/functoria_tool.ml
@@ -321,24 +321,23 @@ module Make (Config: Functoria_sigs.CONFIG) = struct
       ~sdocs:global_option_section
       ~doc
       ~man
-
-  let commands () = [
-    configure ();
-    describe ();
-    build ();
-    clean ();
-    help;
-  ]
-
-  let () =
-    try match Term.eval_choice ~catch:false default (commands ()) with
-      | `Error _ -> exit 1
-      | _ -> ()
-    with
-    | Functoria_misc.Log.Fatal s ->
-      Log.show_error "%s" s ;
-      exit 1
 end
 
 let initialize (module Config:Functoria_sigs.CONFIG) =
- let module M = Make(Config) in ()
+  let module M = Make(Config) in
+  let open M in 
+  try
+    let commands = [
+      configure ();
+      describe ();
+      build ();
+      clean ();
+      help;
+    ] in
+    match Term.eval_choice ~catch:false default commands with
+      | `Error _ -> exit 1
+      | _ -> ()
+  with
+  | Functoria_misc.Log.Fatal s ->
+    Log.show_error "%s" s ;
+    exit 1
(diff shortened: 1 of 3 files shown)


Drup:
I really appreciate the effort, unfortunately, you untangled only the part that was not very tangled. :)
Also, you moved the subcommand info far away from the function they launch, while they are strongly related.
If other people find the code clearer that way, why not. 

95% of the issues are in with_config.

app/functoria_tool.ml
```diff
@@ -18,102 +18,186 @@ open Cmdliner
...
+  Arg.(value & opt (some file) None & doc)
+
+let output =
+  let doc =
+    Arg.info ~docs:global_option_section ~docv:"FILE" ["o"; "output"]
+      ~doc:"File where to output description or dot representation."
+  in
+  Arg.(value & opt (some string) None & doc)
+
+let color =
```
dbuenzli: If you want to remove a few lines you can use [`Fmt_cli.style_renderer`](http://erratique.ch/software/fmt/doc/Fmt_cli.html#VALstyle_renderer) available from `fmt` 0.7.1 (you will need to add a constraint in the `opam` file)

yallop:
Indeed: this is just a start.  More untangling to follow!  I prefer to merge uncontroversial changes first so that when the interesting changes arrive later they're small enough to focus on easily.

And I agree about `with_config`.  The combination of applicatives (with an escape hatch), laziness and side effects doesn't naturally lend itself to being easy to understand.

app/functoria_tool.ml
yallop: Thanks -- I'll take a look.

Drup:
@yallop great :) 

I'm familiar with the code enough that both version are equally readable to me, so I'll let someone else take an additional look. @samoht ?

samoht:
LGTM. Thanks @yallop!

yallop:
Thanks!  It's ok to leave this open for a while, if you prefer, and I'll push more commits to the branch later.

samoht:
Let's merge the simple changes first then :-)

PR #44, opened 2015-12-05
Solve multi-connect issues

Drup:
This is a proposal to solve https://github.com/mirage/mirage/issues/476 and all the other related issues.
Currently, functoria emits a unit-function to delay execution. This assumes referential transparency, which is completely wrong in the context of connect functions. Using lazy should solve the issue.
@yomimono tested under xen and it seems to work.

It breaks https://github.com/mirage/mirage/pull/466
I will try come up with a good idea on how to fix that, otherwise, there is an ugly fix [here](https://github.com/Drup/mirage/commit/5e92bef72a1a7672e95b15f4e9d1b800892ab7e1).

I think this is a good solution, but there might be an Lwt-only solution that doesn't involve lazyness. Ideas ?

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -232,7 +232,7 @@ module Engine = struct
     tbl
 
   let meta_init fmt (connect_name, result_name) =
-    Fmt.pf fmt "let _%s =@[@ %s () @]in@ " result_name connect_name
+    Fmt.pf fmt "let _%s =@[@ Lazy.force %s @]in@ " result_name connect_name
 
   let meta_connect error fmt (connect_name, result_name) =
     Fmt.pf fmt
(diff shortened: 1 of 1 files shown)


Drup:
[Gist of the main.ml for the network example](
It breaks https://github.com/mirage/mirage/pull/466
I will try come up with a good idea on how to fix that, otherwise, there is an ugly fix [here](https://github.com/Drup/mirage/commit/5e92bef72a1a7672e95b15f4e9d1b800892ab7e1).
).

Drup:
Important note: this doesn't protect you from the situation where you create two netif devices with two keys and pass the same interface on the command line.

Drup:
@samoht ? Others ?

samoht:
I'm merging this, this looks ok to me.

PR #46, opened 2015-12-15
Functoria tool refactoring

yallop:
### Summary 

As [promised](https://github.com/mirage/functoria/pull/43#issuecomment-162024497), this PR simplifies the implementation of Functoria's command-line parsing.

It includes lots of commits, each of which makes a mostly-trivial self-contained change (e.g. 2aca45c7a8, a9d04b3ba, 6098877c).  Let me know if you'd prefer them to be squashed.

### Interface changes

There are two minor changes to the external interface: `Functoria_key.with_deps` is deprecated, and there is a new, more primitve function, `Functoria_key.of_deps`.  More on that in a moment.

There are more significant changes to the internal interface: the `Functoria_tool` module has become [`Functoria_command_line`](https://github.com/yallop/functoria/blob/d468cd696ade2c145a2ca4ad734e5c8e881ddaea/app/functoria_command_line.mli), and the interface has been completely redesigned.  The old [`Functoria_tool`](https://github.com/mirage/functoria/blob/master/app/functoria_tool.mli) exposed a single function which accepted a module of functions with side effects, which it called during command-line parsing.  The new `Functoria_command_line` includes five simple functions, each of which parses the command line, performs no effects (except for printing help messages) and returns a value representing the result of parsing:

* The [`read_log_level`](https://github.com/yallop/functoria/blob/d468cd696ade2c145a2ca4ad734e5c8e881ddaea/app/functoria_command_line.mli#L19-L26), [`read_colour_option`](https://github.com/yallop/functoria/blob/d468cd696ade2c145a2ca4ad734e5c8e881ddaea/app/functoria_command_line.mli#L28-L34), [`read_config_file`](https://github.com/yallop/functoria/blob/d468cd696ade2c145a2ca4ad734e5c8e881ddaea/app/functoria_command_line.mli#L36-L38) and [`read_full_eval`](https://github.com/yallop/functoria/blob/d468cd696ade2c145a2ca4ad734e5c8e881ddaea/app/functoria_command_line.mli#L40-L42) functions read a single option from the command line.  The corresponding options need to be read before the main parsing phase -- for example, the config file specified modified the set of available command line options -- and having separate functions to read them makes it easy to do so.

* The [`parse_args`](https://github.com/yallop/functoria/blob/d468cd696ade2c145a2ca4ad734e5c8e881ddaea/app/functoria_command_line.mli#L81-L126) function is the main entry point.  Calling `parse_args` parses the command line (passed as argument) and returns an [`action`](https://github.com/yallop/functoria/blob/d468cd696ade2c145a2ca4ad734e5c8e881ddaea/app/functoria_command_line.mli#L70-L75) value which represents the subcommand and its arguments.  The intention is that the caller should inspect the return value and take appropriate action; `parse_args` does not take any action itself, except for printing help messages when appropriate.

### Implementation changes

The previous implementation was difficult to modify, since the control flow was quite intricate, and involved callbacks, laziness, functors, and communication via global references.  The new implementation is intended to be easy to modify: it builds `Cmdliner` values to represent the command-line argument specifications, and uses them to parse the command line (using `Cmdliner`'s `eval_*` functions) and build values representing the results.

### Tests

There are now some [tests](https://github.com/yallop/functoria/blob/d468cd696ade2c145a2ca4ad734e5c8e881ddaea/tests/test_functoria_command_line.ml) for command-line parsing.  They're currently fairly minimal -- for example, there's little-to-no coverage of failure cases -- but perhaps they'll inspire further testing.  Along with the tests comes a test-time dependency on `ounit`.

### with_deps / of_deps

As I said above, [`with_deps`](https://github.com/mirage/functoria/blob/6ce01159c54fef3a333f0f0566fb6c30bd45a713/lib/functoria_key.mli#L195-L196)

```ocaml
val with_deps : t list -> 'a value -> 'a value
```

is now deprecated, and there's a [new function](https://github.com/yallop/functoria/blob/d468cd696ade2c145a2ca4ad734e5c8e881ddaea/lib/functoria_key.mli#L195-L196),

```ocaml
val of_deps : Set.t -> unit value
```

The new function is more primitive: it can be used to define `with_deps` like this:

```
let with_deps keys v = pure (fun x _ -> x) $ v $ of_deps (Set.of_list keys)
```

The reason for the change is that applicatives for a type `t` generally work better where each function in the interface has type

```ocaml
a1 -> a2 -> ... -> an t
```

and where `t` does not occur in any of the `ai`.  One practical reason for this is that the applicative laws, which make it possible to refactor all applicative expressions into a simple normal form, only apply to functions of this form.  If you have some code like this:

```ocaml
let g =
  let f e2 =
    pure (fun x y -> f (x,y))
      $ e1
      $ e2
   in pure (fun a b -> g (a, b))
        $ e3
        $ e4
```

then you can use the applicative laws to rewrite the code into a form with a single `pure` and no nesting of `$`:

```ocaml
let g =
  pure (fun x a b -> f (x,g (a, b)))
    $ e1
    $ e3
    $ e4
```

However, in code that uses a function like `with_deps`, the argument to the function gets "stuck inside", and the applicative laws don't offer any way to simplify the expression:

```ocaml
let g =
  let f e2 =
    pure (fun x y -> f (x,y))
      $ e1
      $ e2
   in with_deps (pure (fun a b -> g (a, b))
                   $ e3)
```

(The particular change introduced in this pull request allows even more refactorings than usual, since `of_deps` commutes with other "effects", so you change `pure f $ x $ of_deps d` into `pure f $ of_deps d $ x`, and since adjacent `of_deps` effects can be coalesced, so you can turn `pure f $ of_deps d1 $ of_deps d2` into `pure f $ of_deps (union d1 d2)`.)


yallop:
This code isn't heavily tested, so a review would be worthwhile, since the changes are fairly extensive.  I've checked that it passes the tests, that it doesn't change the behaviour on `mirage-skeleton`, and that it seems to behave reasonably at the top level.

yallop:
I also have a [fix](https://github.com/yallop/functoria/compare/functoria_tool-refactoring...yallop:version-help-fix) for mirage/mirage#472.  I'll open another PR if/when this one is merged.

amirmc:
Your comment is worthy of being a blog post in its own right. Thanks for the detailed summary! :+1:

If I understood correctly, you're seeking a code review from mirage *devs*. Am I right?  I'm asking because I wondered if comments from a user perspective would also be helpful (though it might be too soon).  I'm thinking of Matt Gray who's used the current version of Functoria.

*(PS: I'm not really trying to draw a big distinction between users and devs here, just using it for illustration)*

yallop:
Reviews from anyone are welcome!  If Matt and other users of functoria have an opportunity to pin this branch and check that nothing breaks, that'd be helpful:

```
opam pin add functoria https://github.com/yallop/functoria.git#functoria_tool-refactoring
```

(The above should cause `mirage` and anything else that depends on `functoria` to be recompiled.)

amirmc:
@mattgray, Since you've previously used the unreleased version of Functoria (from mirage-dev), it would be great if you could try out Jeremy's branch.  Do you have some time to look at this?

samoht:
+100

That looks incredible! I'm merging this, later corrections are still very welcome.

samoht:
After a very quick review, all of this seems ok. I'll report back if I have any issue.

Drup:
I reviewed all of it. First two points:
- This commit log is very nice to review. It's much more fine-grain that what I usually do, and I'm considering adopting that style now.
- I'm much more on board with the organization of the file after this batch than after the previous one!

I like this new organization very much.

Now, view comments:
- We now have "color" and "colour" in the code base, and I find that absolutely hilarious.
- Do you have a reason to avoid `Sys.argv` ? 
- `fatalize_error` should be moved to misc. I think an equivalent function is used somewhere else.
- matching of the result of `peek` is done in a few different ways, do you have a rational of that , and if no, maybe we should change them to always do it the same way.
- You deprecate `with_deps`. Either we keep it as a utility function, or we remove it completely. Your argument about of_deps playing well with applicative is very good, but I'm not sure the other function is not useful. Maybe we can remove it and see how users do things.
- We should probably find good names for the functions with primes.

@samoht I notice that @yallop too uses ocamldoc comments inside .ml files. :]

yallop:
Thanks for the detailed review, @Drup!

> * Do you have a reason to avoid Sys.argv ?

One advantage of threading the `argv` through is easier testing, since the tests can make up a command line to pass in.

> * We should probably find good names for the functions with primes.

Agreed!  I changed some (internal) names to make them unique, which makes moving things around much easier.  I'd quite like to keep names unique within the module, but it should be possible to improve on the primed names.

> * fatalize_error should be moved to misc. I think an equivalent function is used somewhere else.
> * matching of the result of peek is done in a few different ways, do you have a rational of that , and if no, maybe we should change them to always do it the same way.
> * You deprecate with_deps. Either we keep it as a utility function, or we remove it completely. Your argument about of_deps playing well with applicative is very good, but I'm not sure the other function is not useful. Maybe we can remove it and see how users do things.

I've opened another [pull request](https://github.com/mirage/functoria/pull/49) to address these three issues.

mattgray:
I'm having a look at this now, as it's been merged it's just `opam upgrade functoria` right?

yallop:
Right.

samoht:
Had a weird bug yesterday with skeleton/static_website the 'describe' command didn't pick the storage stack

Not sure if it was a glitch or not and I haven't had time to investigate

PR #47, opened 2015-12-16
Support options such as --help and --version without a config file

yallop:
Fixes mirage/mirage#472.

### Before

```
$ mirage help
[ERROR] No configuration file config.ml found.
Please specify the configuration file using -f.
```
```
$ mirage --version
[ERROR] No configuration file config.ml found.
Please specify the configuration file using -f.
```
```
$ mirage help config
[ERROR] No configuration file config.ml found.
Please specify the configuration file using -f.
```

### After

```
$ mirage help
MIRAGE(1)                        Mirage Manual                       MIRAGE(1)



NAME
       mirage - The mirage application builder
[...]
```
```
$ mirage --version
2.6.1
```
```
$ mirage help config
MIRAGE-CONFIGURE(1)              Mirage Manual             MIRAGE-CONFIGURE(1)



NAME
       mirage-configure - Configure a mirage application.
[...]
```

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -548,12 +548,11 @@ module Make (P: S) = struct
     (* This is a hack to allow the implementation of
        [Mirage.get_mode]. Once it is removed, the notion of base context
        should be removed as well. *)
+    exception No_base_key_map
     let base_context_ref = ref None
     let get_base_context () = match !base_context_ref with
       | Some x -> x
-      | None ->
-        invalid_arg "Base key map is not available at this point. Please stop \
-                     messing with functoria's invariants."
+      | None -> raise No_base_key_map
 
     let pp_info (f:('a, Format.formatter, unit) format -> 'a) level info =
       let verbose = Log.get_level () >= level in
@@ -573,10 +572,13 @@ module Make (P: S) = struct
     let root = Cmd.realpath (Filename.dirname file) in
     let file = root / Filename.basename file in
     set_config_file file;
-    compile_and_dynlink file >>= fun () ->
-    registered () >>= fun t ->
-    Log.set_section (Config.name t);
-    Ok t
+    try
+      compile_and_dynlink file >>= fun () ->
+      registered () >>= fun t ->
+      Log.set_section (Config.name t);
+      Ok t
+    with Config'.No_base_key_map ->
+      Error "Access to base key map is unavailable at this point"      
 
   let get_base_context = Config'.get_base_context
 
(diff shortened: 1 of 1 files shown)


Drup:
In the current kv_ro example (not the functoria version).

```
% mirage --help
Fatal error: exception Invalid_argument("Base key map is not available at this point. Please stop messing with functoria's invariants.")
```

yallop:
@Drup: I think the `Base key map` problem is fixed now.

Drup:
Well, it's fixed, but the documentation doesn't work at all anymore for `mirage configure --help`.

Drup:
I think the base key issue comes from one of your earlier refactoring.

yallop:
I've pushed a revised version (c94c90b) in which all of `mirage --help`, `mirage --version`, `mirage configure --help` and `mirage help` do something reasonable regardless of whether there's a config file available, and which doesn't fall over with pre-functoria config files such as `kv_ro`.

app/functoria_app.ml
```diff
@@ -612,8 +614,28 @@ module Make (P: S) = struct
...
+        argv
+    in
+    match result with
+    | `Error _ -> exit 1
+    | `Ok Cmd.Help -> ()
+    | `Ok (Cmd.Configure _ | Cmd.Describe _ | Cmd.Build _ | Cmd.Clean _) ->
+      Functoria_misc.Log.fatal "%s" error
+    | `Version
+    | `Help -> ()
+    with Invalid_argument _ -> assert false
```
Drup: Is that a leftover before you changed it into a proper exception ?

app/functoria_app.ml
```diff
@@ -573,10 +572,13 @@ module Make (P: S) = struct
...
-    registered () >>= fun t ->
-    Log.set_section (Config.name t);
-    Ok t
+    try
+      compile_and_dynlink file >>= fun () ->
+      registered () >>= fun t ->
+      Log.set_section (Config.name t);
+      Ok t
+    with Config'.No_base_key_map ->
+      Error "Access to base key map is unavailable at this point"
```
Drup: Maybe we should specify which "point" we are talking about ? Something like "Access to base key map is not available before configuration. For dynamically configured devices, please use the configurable class." ?

app/functoria_app.ml
yallop: Indeed.  I've removed it ( 5ec12cb).

app/functoria_app.ml
yallop: Good idea: fixed (c0ab272d).

Drup:
Nice! I tested, it works well. Thanks!

PR #48, opened 2015-12-16
Add a build-test field to the opam file.

yallop:
Perhaps this'll cause the tests to run during Travis builds.

### DIFF
--- a/opam
+++ b/opam
@@ -12,8 +12,11 @@ dev-repo:     "https://github.com/mirage/functoria.git"
 license:      "ISC"
 
 build: [
-  ["./configure" "--prefix=%{prefix}%"]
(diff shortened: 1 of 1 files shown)


Drup:
You should use opam filters. `onuit {test]` in depends and `--enable-tests {test}` in the configure.

yallop:
Hmm.  I already have the `{test}` filter in `depends`, and pass `--enable-tests` if `ounit` is available, which seems to be [sufficient](https://travis-ci.org/mirage/functoria/jobs/97225540#L1146) to cause the tests to run on Travis.

Drup:
Ah, that's the line I was looking for, and I missed it. Perfect then.

Drup:
The CI failed and I don't understand why: https://travis-ci.org/mirage/functoria/jobs/97356688#L2622. Any idea ?

PR #49, opened 2015-12-16
Fix some issues raised in @Drup's review of #46.

yallop:
https://github.com/mirage/functoria/pull/46#issuecomment-165178237

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -637,7 +637,7 @@ module Make (P: S) = struct
     let context_args = Key.context ~stage:`Configure ~with_required:false config_keys in
     let context = 
       match Cmdliner.Term.eval_peek_opts ~argv context_args with
-      | Some context, _ -> context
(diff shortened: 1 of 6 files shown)


Drup:
There is also the matching of peek in functoria_app, [here](https://github.com/mirage/functoria/blob/a6de812f552c55ad6f5842e561932521a8a27ede/app/functoria_app.ml#L641-L646).

yallop:
That one seems a bit different: it needs the first component of the pair returned by `eval_peek_opts`, whereas the other cases need the second component.

Drup:
I'm pretty sure I wondered why at some point, found the answer, and forgot it again. Which mean we should add a comment. :)

yallop:
I've changed my mind: I don't think there's a difference in behaviour compared to the usual style here, so I've changed the code instead (cdb9ff7).

Drup:
Ok. Holidays have ended, merging!

PR #56, opened 2016-04-06
Add an option to ignore command-line arguments

samoht:
Initial patch is from @Drup

See https://github.com/mirage/mirage/pull/497


samoht:
Done properly with the `init` hooks added in #58

PR #57, opened 2016-04-11
Add connect_raw to avoid forcing inputs

talex5:
This is just the logging changes from https://github.com/mirage/functoria/pull/55, as requested in last week's IRC discussion.

For logging devices, we want to initialise as early as possible. With this change, a logging device can use `connect_raw` instead of `connect` to get the unforced inputs (typically, any devices it needs to
operate plus the unikernel to be monitored). It can then force these in the right order.

Notes:

- `impl` now accepts subclasses so that this patch does not require any changes to mirage. The `connect` methods in mirage should ideally be marked as `private` at some point.

- This API exposes the ugly `error` argument in `connect_raw`. This will go away once the rest of the original PR is merged.

- For testing, the https://github.com/talex5/mirage/tree/logging branch uses this to provide a `with_mirage_logs` function. We agreed that this API should change (so that unikernels get logging by default). However, I don't believe this will require any further changes to functoria.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -42,7 +42,7 @@ let sys_argv = impl @@ object
     method ty = argv
     method name = "argv"
     method module_name = "Sys"
-    method !connect _info _m _ =
+    method! private connect _info _m _ =
       "return (`Ok Sys.argv)"
   end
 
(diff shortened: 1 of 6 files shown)


app/functoria_app.ml
```diff
@@ -279,8 +253,12 @@ module Engine = struct
         let modname = Graph.Tbl.find modtbl v in
         Graph.Tbl.add tbl v ident;
         let names = List.map (Graph.Tbl.find tbl) (args @ deps) in
-        Codegen.append_main "%a"
-          emit_connect (error, ident, names, c#connect info modname)
+        Codegen.append_main
+          "@[<v 2>let %s = lazy (@ \
+           %t@ \
+           )@]@."
+          ident
```
Drup: I would prefer this to be kept in a separated function, like it was before.

lib/functoria.mli
```diff
@@ -200,11 +212,11 @@ class type ['ty] configurable = object
...
-  method connect: Info.t -> string -> string list -> string
-  (** [connect info mod args] is the code to execute in order to
-      initialize the state associated with the module [mod] (usually
-      calling [mod.connect]) with the arguments [args], in the context
-      of the project information [info]. *)
+  method connect_raw: error:(string -> string) -> names:string list -> info:Info.t -> modname:string -> Format.formatter -> unit
+  (** [connect_raw ~error ~names ~info ~modname fmt] writes to [fmt] code
+      to construct an instance of the module. Normally, this means forcing
+      each input in [names], waiting for them all to finish, and calling
+      [connect]. *)
```
Drup: Could you add mention to the `meta` functions above ?

app/functoria_app.ml
talex5: OK, I've moved it back out.

lib/functoria.mli
talex5: Good idea. Done.

Drup:
As discussed during last mirage call, we can merge this for now, until a cleaner solution is implemented (See the various other threads for details).

PR #58, opened 2016-04-27
Init

samoht:
Reverting a commit not needed to implement the logging. Adding an other one not really needed as well, but there is an issue about that anyway.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -383,7 +383,6 @@ module type S = sig
   val name: string
   val version: string
   val driver_error: string -> string
-  val init: job impl list
   val create: job impl list -> job impl
 end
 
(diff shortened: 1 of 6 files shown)


app/functoria_app.ml
```diff
@@ -396,7 +396,8 @@ module Make (P: S) = struct
   let register ?(packages=[]) ?(libraries=[]) ?keys:ckeys name jobs =
     let ckeys = match ckeys with None -> [] | Some x -> x in
     let root = get_root () in
-    let main_dev = P.create (keys P.argv :: jobs) in
+    (* we don't have control oover the list of [jobs] *)
+    let main_dev = P.create (P.init @ jobs) in
```
Drup: Hum, can you explain ?

PR #62, opened 2016-06-30
default module_name to Functoria_runtime

hannesm:


hannesm:
(motivation/discussion via https://github.com/mirage/mirage-skeleton/pull/147#discussion_r69032625)

Drup:
I'm not very willing to have a default name there. I want it to be a conscious choice to put a dummy name.

Most of the time, you really want a module name and it would be harmful to forget it.

samoht:
Putting a dummy name is not user-friendly at all. Most of the time the combinator are very simple (at least the combinators that people have started to write) and adding a dummy name is just confusing (at least it confused me multiple time already, I had to look at the code to understand that's it is indeed dummy). I am still not sold on exposing that interface to our users TBH.

Drup:
Maybe a good solution would be to provide a combinator that takes only the configure and clean functions ?

hannesm:
@Drup I like your proposed solution to provide a combinator for configuration steps that only need to execute some shell commands, instead of this PR.

yomimono:
I like the idea of a combinator which just has executes some configuration code much better, as well. Much less complex and closer to what the user may be trying to accomplish.

PR #68, opened 2016-09-03
expose Functoria_key.name, and use it to generate a list of runtime k…

yomimono:
…ey names

In order to know the list of expected command line arguments at runtime from `main.ml`, provide an exposed list in `key_gen.ml`.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -60,7 +60,9 @@ module Keys = struct
     Fmt.pf fmt "@[<v>%a@]@." (Fmt.iter Key.Set.iter pp_var) keys;
     let runvars = Key.Set.elements (Key.filter_stage `Run keys) in
     let pp_runvar ppf v = Fmt.pf ppf "%s_t" (Key.ocaml_name v) in
-    Codegen.append fmt "let runtime_keys = %a" Fmt.Dump.(list pp_runvar) runvars;
+    let pp_names ppf v = Fmt.pf ppf "%S" (Key.name v) in
+    Codegen.append fmt "let runtime_keys = List.combine %a %a"
+      Fmt.Dump.(list pp_runvar) runvars Fmt.Dump.(list pp_names) runvars;
     Codegen.newline fmt;
     R.ok ()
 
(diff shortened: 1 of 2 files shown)


Drup:
Wouldn't you rather want to emit an association list, rather than two lists ?

app/functoria_app.ml
```diff
@@ -61,6 +61,8 @@ module Keys = struct
     Fmt.pf fmt "@[<v>%a@]@." (Fmt.iter Key.Set.iter pp_var) keys;
     let runvars = Key.Set.elements (Key.filter_stage `Run keys) in
     let pp_runvar ppf v = Fmt.pf ppf "%s_t" (Key.ocaml_name v) in
+    let pp_names ppf v = Fmt.pf ppf "\"%s\"" (Key.name v) in
```
Drup: `%S` ;)

yomimono:
> Wouldn't you want to emit an association list, rather than two lists ?

For the use case I have in mind, we'd just be unzipping an association list to get at the first member anyway, but I don't have strong feelings about it.  Can you elaborate a bit?

Drup:
> Can you elaborate a bit?

It's better to directly provide a datatype that shows that the list are of equal lenght, rather than making the user figure it out.

yomimono:
Very delayed thanks for the feedback @Drup - I'm looking at the larger change that inspired this again and I've done as you suggested.  Please merge if you think it's OK for inclusion.

Drup:
Look good to me!

Drup:
Do we need to change documentation somewhere for this?

yomimono:
Hm, `grep`ping around for `runtime_keys` didn't return me anything other than this Codegen usage, including docstrings anywhere.  The only `runtime_keys` usage I turn up in Mirage is the one I added myself in an unmerged branch.  If I'm missing anything, I'm happy to add it, but those are the only places I can think to look.

Drup:
Fair enough, it should probably be documented, but we can add it later.

yomimono:
Thanks, @drup! :)

PR #71, opened 2016-10-01
connect returns no error

hannesm:


Drup:
The reason it's `return` and no `Lwt.return` everywhere is that `functoria` doesn't depend on Lwt, and we expect a prelude provided by the application. We could revisit that, but it doesn't seem necessary here.

hannesm:
@Drup maybe functoria does not _explicitly_ depend on Lwt, but certainly the current master (a3c5d3b) already emits `Lwt.return` sometimes (see e.g. https://github.com/mirage/functoria/blob/a3c5d3b6609d98a03eba114eb4f0edd03e4024c6/lib/functoria.ml#L133)

I think we have to be honest that functoria does actually implicitly plays well with Lwt only atm (since nobody tried anything else) -- and I find it much less confusing to emit what we mean.  Additionally, in case you want to use functoria without Lwt, you only have to provide a mockup module `Lwt` which contains some `return`.

Drup:
Hum, I missed that `Lwt.return`, it shouldn't be there.

Fair enough, but in this case, add the dependency, fix the documentation and use `Lwt.fail` instead of `failwith` in the runtime. ;)

hannesm:
OTOH, I think you're right, and the documentation is well written on that part :)  I force-pushed to get rid of `Lwt.`

Issue #74, opened 2016-10-23
fight bitrot in Functoria_misc


hannesm:
~~~it seems Functoria_misc.Set is for `OCaml<4.02` compatibility, and can be removed~~~.  Also, instead of ~~~introducing Functoria_misc.String, astring could be used~~~ (and llikely Functoria_misc.Cmd can be replaced by bos)...

hannesm:
now finished by #84

PR #77, opened 2016-11-06
use Astring instead of String and custom helpers

hannesm:
as well part of #74


Drup:
Can we re-enable warn error on warning 3 with this ?

hannesm:
sure, there you go! :)  thx for the remark

PR #80, opened 2016-11-15
functoria-runtime: new opam package

hannesm:
exclude <runtime> from include in _tags
no need to require (in META) functoria.runtime from functoria.runtime


Drup:
@samoht Any opinion on that one ? I have nothing against, it makes things a bit clearer. The runtime part was made to be completely independent.

PR #82, opened 2016-11-19
[RFC] No opam , version constraints for packages of unikernels

hannesm:
I played around a bit... so instead of having `~packages:string list ~libraries:string list`, why not have a `Package.t` type and use that (``package : ?ocamlfind:[ `Prefix of string list | `Full of string list ] -> ?min:string -> ?max:string -> string -> t``)

turns out, it is possible... but there are layering violations in mirage (expecting that during configure phase packages are already installed).   the goal here is to just generate an `opam` file from `mirage configure`, and let someone else install dependencies before invoking `make`

see https://github.com/mirage/mirage/pull/691

feedback welcome @samoht @Drup @yomimono @avsm

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -82,8 +82,7 @@ let keys (argv: argv impl) = impl @@ object
     method module_name = Key.module_name
     method !configure = Keys.configure
     method !clean = Keys.clean
-    method !libraries = Key.pure [ "functoria-runtime" ]
-    method !packages = Key.pure [ "functoria-runtime" ]
+    method !packages = Key.pure [package "functoria-runtime"]
     method !deps = [ abstract argv ]
     method !connect info modname = function
       | [ argv ] ->
@@ -98,6 +97,8 @@ let keys (argv: argv impl) = impl @@ object
 type info = Info
 let info = Type Info
 
+(* hannes is pretty sure the following pp needs adjustments, but unclear to me
+   what exactly to do here? i.e. who reads the formatted stuff in again? *)
 let pp_libraries fmt l =
   Fmt.pf fmt "[@ %a]"
     Fmt.(iter ~sep:(unit ";@ ") List.iter @@ fmt "%S") l
@@ -113,7 +114,7 @@ let pp_dump_info module_name fmt i =
     "%s.{@ name = %S;@ \
      @[<v 2>packages = %a@]@ ;@ @[<v 2>libraries = %a@]@ }"
     module_name (Info.name i)
-    pp_packages (Info.packages i)
+    pp_packages (Info.package_names i)
     pp_libraries (Info.libraries i)
 
 let app_info ?(type_modname="Functoria_info")  ?(gen_modname="Info_gen") () =
@@ -123,8 +124,7 @@ let app_info ?(type_modname="Functoria_info")  ?(gen_modname="Info_gen") () =
     method name = "info"
     val gen_file = String.Ascii.lowercase gen_modname  ^ ".ml"
     method module_name = gen_modname
-    method !libraries = Key.pure ["functoria-runtime"]
-    method !packages = Key.pure ["functoria-runtime"]
+    method !packages = Key.pure [package "functoria-runtime"]
     method !connect _ modname _ = Fmt.strf "return %s.info" modname
 
     method !clean i =
@@ -159,21 +159,15 @@ module Engine = struct
     | App     -> Key.Set.empty
 
   module M = struct
-    type t = String.Set.t Key.value
-    let union x y = Key.(pure String.Set.union $ x $ y)
-    let empty = Key.pure String.Set.empty
+    type t = package list Key.value
+    let union x y = Key.(pure List.append $ x $ y)
+    let empty = Key.pure []
   end
 
   let packages =
     let open Graph in
     Graph.collect (module M) @@ function
-    | Impl c     -> Key.map String.Set.of_list c#packages
-    | If _ | App -> M.empty
-
-  let libraries =
-    let open Graph in
-    Graph.collect (module M) @@ function
-    | Impl c     -> Key.map String.Set.of_list c#libraries
+    | Impl c     -> c#packages
     | If _ | App -> M.empty
 
   (* Return a unique variable name holding the state of the given
@@ -316,8 +310,7 @@ module Config = struct
   type t = {
     name     : string;
     root     : string;
-    libraries: String.Set.t Key.value;
-    packages: String.Set.t Key.value;
+    packages: package list Key.value;
     keys    : Key.Set.t;
     init    : job impl list;
     jobs    : Graph.t;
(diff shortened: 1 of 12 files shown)


hannesm:
opam output for mirage-console:

```
console] mirage configure -t virtio
opam-version: "1.2"
name: console
tags: ["org:mirage" ; "org:unikernel"]
build: [ make "build" ]
depends: [
  "ocamlbuild" {build}
  "ocamlfind" {build}
  "solo5-kernel-virtio" 
  "mirage-types-lwt" 
  "mirage-types" {<"4.0.0" & >="3.0.0"}
  "mirage-solo5" 
  "mirage-runtime" {<"4.0.0" & >="3.0.0"}
  "mirage-logs" 
  "mirage-console-solo5" 
  "mirage-clock-freestanding" 
  "mirage-bootvar-solo5" 
  "lwt" 
  "functoria-runtime" 
  "duration" 
]
```

avsm:
Woah, awesome!

samoht:
That's awesome, thanks for having a look at this. I am not totally convinced by the `Full / Suffix` stuff but why not. Also, where do you plan to generate the `opam` file? At the root? In a subdir? This has a direct impact on multi-config support.

hannesm:
glad that you point out the crucial bits:
-`Full` vs `Suffix` is a developer hack - I needed to find out what is used out there (what are the common patterns for ocamlfind package names, apart from when opam name and findlib name are equal.  my current guess after getting mirage to compile and parts of mirage-skeleton is that often for a package foo, `use foo and foo.bar`, but there's also the case (looking at you, mirage-conduit) where the ocamlfind package is disconnected from the opam package.  I guess two keyword arguments, `?sub:string -> ?ocamlfind:string list` which are mutually exclusive, would be a good (and flexible) approximation
- the `opam` file name and `name` (inside of opam) of the unikernel should include the relevant configuration arguments (`console-virtio.opam`, `static_website_tls-socket-unix.opam`), open for suggestions

hannesm:
this has been updated, and I think it is more or less good now... I don't know much about the pp in Functoria_app (and what it should print now, see comment), and also don't have a clue whether we need better errors (and in which way) instead of invalid_arg if the version number is not a list of numbers etc...

I addressed @samoht concerns: (changed to `package : ?sub:string -> ?ocamlfind:string list -> ?min:string -> ?max:string -> string -> package`); and providing a `?name:string` for the opam printer -- thus it should be the concern of the caller ;)

feedback welcome!

samoht:
I really like generating random names for the opam files (when it is not provided by the user). I find it quite nice in [docker](https://github.com/docker/docker/blob/master/pkg/namesgenerator/names-generator.go). And I really want to have my`spitting_lama` unikernel :p

hannesm:
instead of random, why not deterministic (hash over all the provided configuration keys?)

hannesm:
I updated the mirage bits https://github.com/mirage/mirage/pull/691, and https://github.com/mirage/mirage-skeleton/pull/198

will not compile (unless using all these branches and rewriting the CI scripts slightly)

hannesm:
is better now, still questions are open: what should `Functoria_app.pp_dump_info` do?  how should error handling be done (throwing from `Functoria.package` (and merge, etc.) seems a bit rough)?

hannesm:
@Drup I'm interested in your opinion here (and maybe you have answers to the 2 questions I raised in my previous comment).

lib/functoria.ml
```diff
@@ -21,39 +21,147 @@ open Functoria_misc
...
-    { name; root; keys; libraries; packages; context }
-
-  let pp verbose ppf { name ; root ; keys ; context ; libraries ; packages } =
+    let packages = List.fold_left (fun m p ->
+        let n = p.opam in
+        match String.Map.find p.opam m with
+        | None -> String.Map.add n p m
+        | Some p' -> match Package.merge p.opam p p' with
+          | Some p -> String.Map.add n p m
+          | None -> invalid_arg "bad constraints")
```
Drup: We should provide a better error message (and in particular, which opam package triggers the conflict).

lib/functoria.ml
```diff
@@ -21,39 +21,147 @@ open Functoria_misc
...
+  let libraries t =
+    String.Set.elements
+      (List.fold_left String.Set.union String.Set.empty
+         (List.map (fun x -> x.ocamlfind)
+            (List.filter (fun x -> x.build = false)
+               (List.map snd (String.Map.bindings t.packages)))))
+  let package_names t =
+    List.map fst
+      (List.filter (fun (_, p) -> p.build = false)
+         (String.Map.bindings t.packages))
```
Drup: I feel like you should move that stuff to the `Package` module, and keep only `Info.package`.

app/functoria_app.ml
```diff
@@ -113,7 +114,7 @@ let pp_dump_info module_name fmt i =
     "%s.{@ name = %S;@ \
      @[<v 2>packages = %a@]@ ;@ @[<v 2>libraries = %a@]@ }"
     module_name (Info.name i)
-    pp_packages (Info.packages i)
+    pp_packages (Info.package_names i)
```
Drup: `pp_dump_info` is used to expose the content of `Info` to the underlying unikernel, in order to do things like an unikernel that lists the packages and the version it uses. It dumps an ml files with placeholder like `%{functoria:version}%` and then call `opam config subst` to substitute them.

That probably won't really work well anymore, so we need to move `opam config subst` to the "make depend" target.
Since you have a bit more information now, we could enrich the datastructure in functoria-runtime with packages and dump that.

lib/functoria.ml
hannesm: no.  The Info record keeps track of a compilation unit, which contain a map of packages.  The package module does only care about a single package.

Drup: Meh, `Package` is the module that deals with packages. You want something from `package list` to the list of ocamlfind (resp. opam packages), I feel like it would belong there.

hannesm:
this is now pretty-printing a bit prettier:

```
$ less mirage-unikernel-console-virtio.opam
# Generated by console (Mon, 21 Nov 2016 17:03:06 GMT).
opam-version: "1.2"
name: "mirage-unikernel-console-virtio"
depends: [ "duration" 
           "functoria-runtime" 
           "lwt" 
           "mirage-bootvar-solo5" 
           "mirage-clock-freestanding" 
           "mirage-console-solo5" 
           "mirage-logs" 
           "mirage-runtime" {>="3.0.0"}
           "mirage-solo5" 
           "mirage-types" {>="3.0.0"}
           "mirage-types-lwt" 
           "ocamlbuild" {build}
           "ocamlfind" {build}
           "solo5-kernel-virtio" 
]
```

```
$ less mirage-unikernel-console-xen.opam 
# Generated by console (Mon, 21 Nov 2016 17:06:27 GMT).
opam-version: "1.2"
name: "mirage-unikernel-console-xen"
depends: [ "duration" 
           "functoria-runtime" 
           "lwt" 
           "mirage-bootvar-xen" 
           "mirage-clock-freestanding" 
           "mirage-console-xen" 
           "mirage-logs" 
           "mirage-runtime" {>="3.0.0"}
           "mirage-types" {>="3.0.0"}
           "mirage-types-lwt" 
           "mirage-xen" 
           "ocamlbuild" {build}
           "ocamlfind" {build}
]
```

```
$ less mirage-unikernel-console-unix.opam 
# Generated by console (Mon, 21 Nov 2016 17:06:50 GMT).
opam-version: "1.2"
name: "mirage-unikernel-console-unix"
depends: [ "duration" 
           "functoria-runtime" 
           "lwt" 
           "mirage-clock-unix" 
           "mirage-console-unix" 
           "mirage-logs" 
           "mirage-runtime" {>="3.0.0"}
           "mirage-types" {>="3.0.0"}
           "mirage-types-lwt" 
           "mirage-unix" 
           "ocamlbuild" {build}
           "ocamlfind" {build}
]
```

thing left is the info stuff drup explained above... this will need more thinking and integration into the build phase (but i guess it is ok to have it partially broken for some time)...

Drup:
I'm pretty happy with the last batch of changes. We just need to fix `pp_dump_info`/`app_info`.

hannesm:
superseeded by #84

Issue #81, opened 2016-11-16
`describe --dot` opens xdot but doesn't display the graph


yomimono:
If I save the output with `-o my-rad-graph.dot` and then open with `xdot my-rad-graph.dot`, I see the graph as expected.

Drup:
functoria pipes the content of the dot to `xdot` directly. It works on my system but not on @yomimono's. Not sure why. I suppose we should create a temporary dot file and give it to `xdot`.

PR #76, opened 2016-11-04
Removing the build subcommand

hannesm:
this was legacy from before functoria, MirageOS instructions contain
`mirage configure` followed by `make` -- there's no need for this indirection

Issue #63, opened 2016-06-30
Provide a way to run commands without escaping issues


talex5:
https://github.com/mirage/mirage-skeleton/pull/147 does:

```
Functoria_app.Cmd.run "rm -f %s/disk.img" dir
```

This is unsafe if `dir` contains spaces or other special characters. Functoria should provide an alternative that doesn't go via the shell.

`Lwt_process` and `Unix.create_process` take arrays instead.
`Bos.Cmd` makes command-lines abstract to acheive the same thing:
https://github.com/dbuenzli/bos/blob/78bd66169f401e6ebf6347cd903cc2f1c3f0b7f8/src/bos.mli#L115

samoht:
+1 to use [Bos](https://github.com/dbuenzli/bos) as much as possible now that it is released :-)

hannesm:
we now use bos, there is no longer any `Functoria_app.Cmd`... bos does it right, you can use it in your own configure and build

Issue #54, opened 2016-03-20
Fix quoting of opam arguments


samoht:
See https://github.com/mirage/mirage/pull/519

samoht:
There seem to be a bug in `depext` which doesn't support `>=` constraints anyway, so will work-around that bug by just removing the constraint...

```
     method packages =
 +        (* FIXME: fix the bug in functoria upstream *)
 +        let quote s = Printf.sprintf "'%s'" s in
          let l = [
 -          "lwt"; "mirage-types" ^ mirage_types_version; "mirage-types-lwt"
 +          "lwt"; quote ("mirage-types" ^ mirage_types_version);
 +          "mirage-types-lwt"
          ] in
          Key.(if_ is_xen) ("mirage-xen" :: l) ("mirage-unix" :: l)
```

hannesm:
solved by #84

PR #84, opened 2016-11-25
New order

### LINKED ISSUE #73: custom logs, replace with logs library
functoria_app has a custom logs signature and implementation, which should be replaced by the fine logs library, code: https://github.com/mirage/functoria/blob/d6a3bd3671c4c8645ad0ec2fb224affac3914e39/app/functoria_app.mli#L166

hannesm:
This is on top of #82, but also removes Cmd and Log, replacing it with Bos and Logs.  ~~~For some unknown reason `mirage clean` does not work as expected here. I might look into that later... otherwise, comments/feedback welcome (I'm sure I missed something)~~~

It also re-introduces the `build` command (which calls out to `bla#build` in the graph ('coz @samoht is always right ;) (reverts #76)

Fixes #73 #74 #63 (removing Cmd) #54 (superseeds #82) #81

ready to go, commits can be squashed imho.


app/functoria_app.ml
```diff
@@ -417,81 +435,113 @@ module Make (P: S) = struct
...
+    | Error _ -> R.error_msg "failed to change directory for configure"
+
+  let build i jobs =
+    Log.info (fun m -> m "Building: %s" (get_config_file ()));
+    match
+      Bos.OS.Dir.with_current (Fpath.v (Info.root i))
+        (fun () -> Engine.build i jobs) ()
+    with
+    | Ok a -> a
+    | Error _ -> R.error_msg "failed to change directory for build"
```
Drup: Could you move the logic for directory changing in an auxiliary function ?

app/functoria_app.ml
```diff
@@ -417,81 +435,113 @@ module Make (P: S) = struct
...
-          root root
-      )
+    match
+      Bos.OS.Dir.with_current (Fpath.v root)
+        (fun () ->
+           clean_main i job >>= fun () ->
+           Bos.OS.Dir.delete ~recurse:true (Fpath.v "_build") >>= fun () ->
+           Bos.OS.File.delete (Fpath.v "log") >>= fun () ->
+           Bos.OS.File.delete (Fpath.v "main.native.o") >>= fun () ->
+           Bos.OS.File.delete (Fpath.v "main.native")) ()
```
hannesm: was the same before https://github.com/mirage/functoria/blob/d7a20157078e1f1e9b2680ce24031e44d19faaa3/app/functoria_app.ml#L488

since this PR is already pretty big, I'd leave the `main.native` question to a subsequent PR.

app/functoria_app.ml
```diff
@@ -417,81 +435,113 @@ module Make (P: S) = struct
...
     let with_fmt f = match output with
       | None when dot ->
-        let f oc = Cmd.with_channel oc f in
-        Cmd.with_process_out dotcmd f
-      | None -> f Fmt.stdout
-      | Some s -> Cmd.with_file s f
-    in R.ok @@ with_fmt f
+        f Format.str_formatter ;
+        let data = Format.flush_str_formatter () in
+        Bos.OS.Cmd.run_in (Bos.Cmd.v dotcmd) (Bos.OS.Cmd.in_string data)
```
Drup: While you're at it (and you seem fluent in `Bos`), could you pipe that to a temp file and call `dotcmd` on the tempfile, instead ? See #81.

hannesm: you mean `pp > tmp` followed by `dotcmd tmp`?

lib/functoria.mli
```diff
@@ -235,14 +235,19 @@ class type ['ty] configurable = object
...
-  method clean: Info.t -> (unit, string) Rresult.result
+  method build: Info.t -> (unit, Rresult.R.msg) Rresult.result
+  (** [build info] is the code to execute in order to build
+      the device. This might involve generating OCaml code,
+      running shell scripts, etc. *)
+
+  method clean: Info.t -> (unit, Rresult.R.msg) Rresult.result
   (** [clean info] is the code to clean-up what has been generated
-      by {!configure}. *)
+      by {!build}. *)
```
Drup: Not both ? How do we clean up things generated by configure then ?

lib/functoria_misc.ml
```diff
@@ -293,7 +83,7 @@ module Codegen = struct
         days.(time.Unix.tm_wday) time.Unix.tm_mday
         months.(time.Unix.tm_mon) (time.Unix.tm_year+1900)
         time.Unix.tm_hour time.Unix.tm_min time.Unix.tm_sec in
-    Format.sprintf "Generated by %s (%s)." name date
+    Format.sprintf "Generated by functoria (%s)." date
```
Drup: I was asked specifically not to put functoria here. The actual application (aka mirage, or the unikernel) should put its name here. Not sure how @samoht 's position has evolved on the matter. :)

app/functoria_app.ml
Drup: Yes.

app/functoria_app.ml
Drup: Ok, fair enough.

Drup:
Not sure why you made a new PR instead on continuing on the old one. It would have been easier to review.

Did you made sure that the log level option appears in the help ? It's not so clear to me by reading the code.

Looks good otherwise.

lib/functoria.mli
```diff
@@ -192,28 +225,29 @@ class type ['ty] configurable = object
...
+  method configure: Info.t -> (unit, Rresult.R.msg) Rresult.result
   (** [configure info] is the code to execute in order to configure
       the device. This might involve generating more OCaml code,
-      running bash scripts, etc. *)
+      running shell scripts, etc. *)
+
+  method build: Info.t -> (unit, Rresult.R.msg) Rresult.result
+  (** [build info] is the code to execute in order to build
+      the device. This might involve generating OCaml code,
+      running shell scripts, etc. *)
```
Drup: There is a clear overlap between the function of configure and build, according to the documentation. It should be much clearer what should go where and when things are executed.

hannesm:
@drup: thanks for your review.  you could read only the very last commit, which is the diff on top of the previous PR.  it was not necessary here, but in the mirage repository a new PR made sense (since it partially reverts previous changes, such as the Makefile pkg-config thingies)

`mirage configure --help` includes:
```
       --color=WHEN (absent=auto)
           Colorize the output. WHEN must be one of `auto', `always' or
           `never'.

       -q, --quiet
           Be quiet. Takes over -v and --verbosity.

       -v, --verbose
           Increase verbosity. Repeatable, but more than twice does not bring
           more.

       --verbosity=LEVEL (absent=warning)
           Be more or less verbose. LEVEL must be one of `quiet', `error',
           `warning', `info' or `debug'. Takes over -v.
```

lib/functoria_misc.ml
samoht: I would still prefer to see `mirage.VERSION` here instead of `functoria` :-)

samoht:
I really like the new generation of opam files, and the cleanup with Bos + Logs is very welcome. I also like the new `build` command but It would be very helpful if you could comment on the new expected build flow. Do you expect the mirage driver in Mirage3 to be able to build one/all of the configured unikernels in one commend? Or something else?

hannesm:
@Drup I (hopefully) fixed all your comments :) (even the main.native)

hannesm:
@samoht I don't understand your question.  I do specifically (atm) not work on the ability to have various configurations (xen/unix/virtio) available at the same time.

The workflow, as mentioned on mirageos-devel, will now be:
`mirage configure -t <backend>`; `make depend`; `mirage build -t <backend>`.

The `make depend` calls out to `opam`, and is only necessary if you don't have all dependencies installed already.

To smooth the transition, a `Makefile` is generated which `build` target contains `mirage build -t <backend>` (but you don't have to have `make` to compile (apart from the ukvm target where `ukvm-configure` generates a `Makefile.ukvm`)).

Thus, for MirageOS unikernel developers: `mirage configure` ; `make` should be sufficient (with subsequent calls to `make` while changing the Caml code).

For CI systems, some instance can `foreach backend ; mirage configure -t $backend`, and then another set of CIs could `opam pin mirage-unikernel-<name>-<backend>` and `mirage build -t <backend>`.  There is no longer a convolution between `configure`, `opam` dependency installations, and `build`.  For incremental rebuilds (reverse dependencies), unless `config.ml` changes, you can use the generated opam files as dependency sources.  Maybe the generated opam file should contain a `build: [ mirage build -t <backend> ]` to have it more streamlined.

lib/functoria_misc.ml
hannesm: I now use `Sys.argv`, this preserves the entire command line.  Not sure how functoria would be able to access the mirage version string.

lib/functoria.mli
```diff
@@ -237,13 +237,15 @@ class type ['ty] configurable = object
...
+      the device.  During the configuration phase, the specficied
+      {!packages} might not yet be there.  The code might involve
+      generating more OCaml code, running shell scripts, etc. *)
 
   method build: Info.t -> (unit, Rresult.R.msg) Rresult.result
   (** [build info] is the code to execute in order to build
-      the device. This might involve generating OCaml code,
-      running shell scripts, etc. *)
+      the device.  During the build phase, you can rely that all
+      {!packages} are available.  This might involve generating
```
Drup: s/available/installed/ 

Maybe mention external tools explicitly ?

Drup:
@hannesm What happens if I call `mirage build` before doing `mirage configure` apart from an unclear error ?

Drup:
> Maybe the generated opam file should contain a build: [ mirage build -t <backend> ] to have it more streamlined.

Yes, that sounds like a good idea.

hannesm:
@drup the opam generated by mirage now emits a build rule.

mato:
When building for the `ukvm` target, I don't see any of the `ukvm` link commands output. Is that intentional?

Also, I tried to get `ocamlbuild` to ignore `_build-ukvm` by placing this in `_tags`:

    <_build_ukvm>: -traverse

but that doesn't seem to help, it still complains when running `make` a 2nd time:

````
$ make
mirage build -t ukvm
ocamlfind ocamlopt unix.cmxa -I /home/mato/.opam/mirage-dev-4.04.0/lib/ocamlbuild /home/mato/.opam/mirage-dev-4.04.0/lib/ocamlbuild/ocamlbuildlib.cmxa -linkpkg myocamlbuild.ml /home/mato/.opam/mirage-dev-4.04.0/lib/ocamlbuild/ocamlbuild.cmx -o myocamlbuild
SANITIZE: a total of 1 file that should probably not be in your source tree
  has been found. A script shell file
  "/home/mato/solo5/mirage-skeleton/stackv4/_build/sanitize.sh" is being
  created. Check this script and run it to remove unwanted files or use other
  options (such as defining hygiene exceptions or using the -no-hygiene
  option).
IMPORTANT: I cannot work with leftover compiled files.
ERROR: Leftover object files:
  File ukvm-core.o in _build-ukvm has suffix .o
Exiting due to hygiene violations.
mirage: unknown option `-t'.
Usage: mirage build [OPTION]... 
Try `mirage build --help' or `mirage --help' for more information.
Fatal error: exception (Invalid_argument
   "run ['ocamlbuild' '-use-ocamlfind' '-classic-display' '-tags'\
  \n     'bin_annot,color(always)' '-pkg' 'mirage' 'config.cmxs']: exited with 1")
Makefile:14: recipe for target 'build' failed
make: *** [build] Error 2
````

I tried the opposite (`<build-ukvm/*>: precious`) but that didn't help either.

mato:
(deleted, wrong PR and typo, sorry)

mato:
@hannesm `mirage build ...` seems to accept all the configure-time options that `mirage configure ...` would for a unikernel (e.g. `--dhcp` ...) That doesn't seem right?

hannesm:
On 28/11/2016 16:07, Martin Lucina wrote:
> @hannesm `mirage build ...` seems to accept all the configure-time options that `mirage configure ...` would for a unikernel (e.g. `--dhcp` ...) That doesn't seem right?

Interesting.  the same is true for `mirage help clean`.  not sure how to
avoid this, but #72 looks very related to this.

PR #85, opened 2016-11-30
preserve arguments across invocation (configure/build/clean)

hannesm:
@mato @yomimono this fixes the static_website_tls with `--dhcp true` by marshalling and unmarshalling the mirage configure arguments.

also now errors out if mirage build (or clean) is called without mirage configure before

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -434,6 +434,14 @@ module Make (P: S) = struct
     | Some t -> Ok t
 
   let configure_main i jobs =
+    Log.info (fun m -> m "Preserving arguments");
+    let args =
+      if Array.length Sys.argv >= 2 then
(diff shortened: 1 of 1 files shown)


mato:
Tested and works as advertised, however the error message on running `mirage build` w/o configuring first could be more user-friendly:
````
Fatal error: exception (Invalid_argument "./.mirage.keys: No such file or directory")
````
I suggest something like "Error: Please run 'mirage configure' before running XXX".

yomimono:
@mato reminded me that you asked for this PR not to be merged on mirageos-devel, but I'd already merged it.  I'd prefer to keep this fix in to make the behavior less surprising and get another patch on top of it, but it's really up to you @hannesm -- we can revert as well.

hannesm:
thanks for merging, this was exactly supposed to be handled as you did -- it improves the current behaviour, but is not a good solution.  thanks for merging, I'll hopefully clean it up in a future PR.

PR #87, opened 2016-12-01
Rewrite configuration caching.

Drup:
Sorry @hannesm, but the original implementation was not reasonable.
In theory, I would prefer to serialize a `Key.context`, but that might be tricky (univ map with exceptions .. I reckon @samoht had issues with that). If we can find a solution, I'll switch to that.

Failure modes should be decent:
- build and clean fails when cache is not present. They don't take any extra arguments
- describe can run without cached conf. If there is a cached conf, it will behave as with --eval. You can still pass extra arguments to see a different configuration.
- configure ignores the cache.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -717,7 +717,11 @@ module Make (P: S) = struct
          Config'.eval ~with_required:true ~partial:false context config
        and describe =
          let context = Cache.merge ~cache:cached_context context in
-         let partial = not (full_eval || Cache.present cached_context) in
+         let partial = match full_eval with
+           | Some true  -> false
+           | Some false -> true
+           | None -> not (Cache.present cached_context)
+         in
          Config'.eval ~with_required:false ~partial context config
        and build =
          Config'.eval_cached ~partial:false cached_context config
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -392,6 +392,70 @@ module Config = struct
 
 end
 
+(** Cached configuration in [.mirage.config].
+    Currently, we cache Sys.argv directly
+*)
+module Cache : sig
+  open Cmdliner
+  val save : string -> (unit, [> Rresult.R.msg ]) result
+  val clean : string -> (unit, [> Rresult.R.msg ]) result
+  val get_context : string -> context Term.t ->
+    [> `Error of bool * string | `Ok of context option ]
+  val require :
+    [< `Error of bool * string | `Ok of context option ] -> context Term.ret
+  val merge :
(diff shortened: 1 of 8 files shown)


samoht:
@Drup the solution is probably https://github.com/let-def/distwit

Drup:
We probably don't want to use marshall for this, though. It's a public file that might be picked up by different ocaml versions, and we want to provide nice errors if we change the format.

Drup:
I just added --no-eval, the complement of --eval, which should satisfy everyone, regarding evaluation of the dot output. :)

samoht:
Yes indeed, Marshal is probably not the answer to the problem that this PR solves, but you were asking about marshalling universal maps  :-)

Can we serialize the keys to JSON maybe?

Drup:
@samoht I'm not sure how to store an `exn StringMap.t` in Json. :/

Drup:
In any cases, It's not urgent, since it's decently abstracted in the implementation, including deserialization failure.

hannesm:
(sorry, was AFK for some time) I thought/hoped we could get away without marshalling the keys, but instead just using the compiled `config.ml` in build (and clean).  I've a half-finished implementation thereof (and wondering whether we need to really recover the keys between invocations).

Drup:
@hannesm That was my initial idea too, unfortunately, we need the keys to redo evaluation if we want `mirage describe` to be able to take extra arguments.

hannesm:
@Drup ah, yes.  another use case for marshalling (and unmarshalling) keys is to have multi-config support (read as `mirage switch`) and a usable UI for that.  thanks for working on this, I'll look once my brain has more power and base my stuff on top of yours!

Drup:
@hannesm Sounds nice! We will need to extend `mirage describe` to show that a configuration (or several!) is cached.

hannesm:
@Drup is this ready to be merged from your side?

Drup:
Yes, but I didn't investigated the build failure.

hannesm:
(testing mirage-dev at https://travis-ci.org/hannesm/mirage-dev/builds/180738513 using @Drup branch)

hannesm:
failures are in the lwt subdir of mirage-skeleton (unfortunately expected), which Makefile and project layout I do not agree with... but we should fix he mirage-skeleton build script to get travis through

yomimono:
Without this change, *all* of the tests are broken because `mirage build` recalls that `mirage configure` was invoked with `-f ${testname}/config.ml`, but since `mirage build`'s `wd` is `${testname}`, it fails to find the `${testname}` subdirectory in there and fails.  I'll merge and then hack on the `lwt` subdir.

PR #88, opened 2016-12-05
Fpath

hannesm:


Drup:
Looks good!

yomimono:
There are a few other places where the error message is "please run configure", which is a bit opaque and probably should be replaced with something more like what's in this PR, but that can be done separately.  Thanks @hannesm! :)

PR #89, opened 2016-12-06
ocamlify opam filenames

yomimono:
Users may call `register` with a name that includes characters that will
be surprising to `opam` (for example, ".").  Use `ocamlify` to restrict
the character set that will be emitted.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -352,6 +352,7 @@ module Config = struct
 
   let make ?(keys=[]) ?(packages=[]) ?(init=[]) name root
       main_dev =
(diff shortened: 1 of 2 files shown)


yomimono:
Merging to get `mirage-www` buildable again, but comments, better fixes, or reverting PRs are welcome.

yomimono:
This PR looks open to me but its commits are reflected in the history for `master`.  I'll close this, which hopefully won't make things any more confusing than they already are...

hannesm:
see #113 (and https://github.com/mirage/mirage/issues/809) for a partial revert.

PR #90, opened 2016-12-06
test ocaml versions 4.03 and 4.04, and not 4.02

yomimono:


Drup:
RIP 4.02!

We can use module aliases now!

PR #91, opened 2016-12-06
restrict -f command line argument to file in cwd

hannesm:
reducing complexity, making mirage and functoria slightly less flexible... now, `config.ml` is required, rather than any ML file and passing `-f`.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -466,20 +466,15 @@ module Make (P: S) = struct
   module Log = (val Logs.src_log src : Logs.LOG)
 
   let configuration = ref None
-  let config_file = ref None
+  let config_file = ref Fpath.(v "config.ml")
 
-  let set_config_file f =
-    config_file := Some f
+  let get_config_file () = !config_file
+  let set_config_file file = config_file := file
 
-  let get_config_file () =
-    match !config_file with
-    | Some f -> f
-    | None ->
-      match Bos.OS.Dir.current () with
-      | Ok p -> Fpath.(p / "config.ml")
-      | Error e -> R.error_msg_to_invalid_arg (Error e)
-
-  let get_root () = Fpath.parent (get_config_file ())
+  let get_root () =
+    match Bos.OS.Dir.current () with
+    | Ok p -> p
+    | Error e -> R.error_msg_to_invalid_arg (Error e)
 
   let register ?packages ?keys ?(init=[]) name jobs =
     let root = get_root () in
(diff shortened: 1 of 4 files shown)


Drup:
I don't disagree (-f is a big headache) but I think you should discuss that on the mailing list to make sure nobody feels strongly about it. :)

samoht:
Removing the recursive search makes sense but I still think that allowing to change the default config filename is useful in some cases (e.g. we should provide good default, but allow local overwrite).

hannesm:
I raised this in the last call, and also on the mailing list.  @samoht could you elaborate on a single case where it is really useful (and how it is not confusing to have either multiple (config.ml, myconfig.ml, myworkingconfig.ml) or a single foo.ml around (and forgetting every 5th time to pass `-f foo.ml`).  I tend to agree on user choices and freedom, but I don't think it is worth the headache (take symlinks, `..`, etc. into account for filename resolution, including switching the working directory for compiling).  OCaml (and opam) have some magic filenames hardcoded (myocamlbuild.ml, _tags, META) - I don't see that config.ml will make this really worth.

But, as @amirmc mentioned in the last call agenda, maybe we should try to get @sgrove attention here, whether they use renamed config.ml and the -f argument for mirage.

samoht:
My rational is that we need a way to scale, in the future, to multiple unikernels per repository (as we today have multiple opam packages per repository). This could only work if you have a way to override the default `config.ml` location. For instance, you could have multiple sub-directories, each with a `config.ml`, or multiple top-level `config.ml`-like files. I am not sure which is the best layout yet, but I want to have a way to keep this choice open.

Removing support for multiple `config.ml` or for the heuristic to find the unique `config.ml` makes complete sense though. Let the user (or the build system) be more explicit about what to build makes complete sense and go in the general direction that we are taking for Mirage3.

hannesm:
@samoht thanks.

Please keep in mind that having a single `config.ml` limits to one unikernel per directory, rather than per repository.

There are multiple things we would like to support:
- multiple configurations of a single unikernel (which build into separate build directories)
- multiple unikernels per repository

We can solve the multiple unikernels per repository by utilising opam: have `N` opam files (where `N = number of unikernels * configuration`) checked in (each has a unique name), containing the concrete build instructions (maybe even the `mirage configure` call?):
`cd <subdir> && make build` (or `cd <subdir> && mirage configure <keys>` followed by `make`).

Using this approach requires to check in generated files (one for each configuration to be tested) -- but these opam files can also be produced elsewhere (and automatically re-generated if either `config.ml` changes or `mirage` is updated).

I'm also ok with having a `-f` flag, if the argument is a filename (without any `/` characters), which would simplify the whole absolute path and root logic (which should be reused for compiling multiple configurations in a single directory (as requested in https://github.com/mirage/mirage/issues/582)).

The current state of functoria master (which may absolutely be my fault since I messed around with those pieces of the code) is that `get_config_file ()` returns the raw argument passed via `-f` (which may be a relative or absolute path), and `get_root ()` returns the parent directory of the configuration (again, may be relative or absolute).  I'd like `get_config_file ()` to return the filename only, and `get_root ()` the absolute path to the working directory.

hannesm:
I updated this PR to reflect my most recent commit.

Functoria has a `-f` `--file` optional argument back (defaults to `config.ml`).  The given FILE must be an existing file with no segments (thus `tests/config.ml` is invalid, but `CHANGES.md` is valid).

This removes the headache around relative and absolute paths (and setting the working dir appropriately), and separates the `root` (cwd) and the `config file` nicely.

@samoht is this ok with you?

hannesm:
@samoht your opinion on this revised PR would be appreciated (supporting `-f` with a filename, but no path).  Thus you have to run mirage in the directory of the unikernel.

samoht:
Yes I guess that's fine. I would prefer to support any path and simplify all of the `get_root`/ `get_config_file` stuff but this could be done later. As long as we keep the `-f` argument for now on (with a weak semantics which can be fixed/extended later), I am fine.

hannesm:
since @samoht mentioned he is fine with this change, and mirage-skeleton builds with this, I'm merging it.  the semantics of `-f` is now much simpler.

Issue #86, opened 2016-12-01
ocamlbuild output is badly handled 


Drup:
1. The command to build `config.ml` are printed by default. They shouldn't, when the build works fine, there is no reason to show anything. It's pure noise
2. When the `config.ml` build fail, I get this:
```
Command exited with code 2.
Fatal error: exception (Invalid_argument
   "run ['ocamlbuild' '-use-ocamlfind' '-classic-display' '-tags'\
  \n     'bin_annot,color(always)' '-X' '_build-ukvm' '-pkg' 'mirage'\
  \n     'config.cmxs']: exited with 10")
```
Which is probably the worst way of reporting errors. :)

Ideally, it should only print the ocaml error. 
@hannesm I would solve it myself, but bos is pretty much symbol non-sense for me at this point (and I don't find the API particularly easy to use).

dbuenzli:
The *what* non-sense ?

hannesm:
fixed https://github.com/mirage/functoria/issues/97

PR #93, opened 2016-12-09
invoke ocamlbuild with quiet

hannesm:
addresses #86, before (with a broken `config.ml`):
```
# mirage configure
Fatal error: exception (Invalid_argument
   "run ['ocamlbuild' '-use-ocamlfind' '-classic-display' '-tags'\
  \n     'bin_annot,color(always)' '-X' '_build-ukvm' '-pkg' 'mirage'\
  \n     'config.cmxs']: exited with 10")
```

after:
```
# mirage configure
+ ocamlfind ocamldep -package mirage -modules config.ml > config.ml.depends
File "config.ml", line 48, characters 0-1:
Error: Syntax error
Command exited with code 2.
Fatal error: exception (Invalid_argument
   "run ['ocamlbuild' '-use-ocamlfind' '-classic-display' '-tags'\
  \n     'bin_annot,color(always)' '-quiet' '-X' '_build-ukvm' '-pkg' 'mirage'\
  \n     'config.cmxs']: exited with 10")
```

and with a good `config.ml`:
```
# mirage configure
#
```

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -570,10 +570,10 @@ module Make (P: S) = struct
       (fun () ->
          let cmd =
            Bos.Cmd.(v "ocamlbuild" % "-use-ocamlfind" % "-classic-display" %
-                    "-tags" % "bin_annot,color(always)" %
+                    "-tags" % "bin_annot,color(always)" % "-quiet" %
                     "-X" % "_build-ukvm" % "-pkg" % P.name % file)
          in
-         Bos.OS.Cmd.(run_out cmd |> to_null) >>= fun _ ->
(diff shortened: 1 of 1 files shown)


Drup:
Maybe we should change the invalid_argument too ?

yomimono:
This is much nicer, @hannesm .  It'd be nice to replace the `invalid_argument` output with some more guided text about a problem compiling the configuration document, but that can be done in a separate patch.

hannesm:
I do agree that the invalid argument error message is of not much use in the current form.

I don't think special casing this _now_ (to not emit an invalid_arg with the ocamlbuild invocation) makes sense, since the argument processing needs another iteration anyways.

PR #97, opened 2016-12-20
fix commandline without --file

hannesm:
if we run mirage without the --file option, `config_file` will be "config.ml"
(which may be non-existing).  Deal with this situation properly.  Still error
out (invalid_arg) if called with a full path (instead of a single segment).

fix mirage with no config file commandline option

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -551,12 +551,9 @@ module Make (P: S) = struct
     in
     with_fmt f
 
-  (* Compile the configuration file and attempt to dynlink it.
-   * It is responsible for registering an application via
-   * [register] in order to have an observable
-   * side effect to this command. *)
-  let compile_and_dynlink file =
-    Log.info (fun m -> m "Processing: %a" Fpath.pp file);
+  (* Compile the configuration file. *)
+  let compile file =
+    Log.info (fun m -> m "Compiling: %a" Fpath.pp file);
     let file = Dynlink.adapt_filename (Fpath.to_string file)
     and cfg = Fpath.rem_ext file
     and root = get_root ()
@@ -570,13 +567,25 @@ module Make (P: S) = struct
                     "-tags" % "bin_annot,color(always)" % "-quiet" %
                     "-X" % "_build-ukvm" % "-pkg" % P.name % file)
          in
-         Bos.OS.Cmd.run cmd >>= fun _ ->
-         try Ok (Dynlink.loadfile Fpath.(to_string (root / "_build" / file)))
-         with Dynlink.Error err ->
-           Log.err (fun m -> m "Error loading config: %s" (Dynlink.error_message err));
-           let msg = Printf.sprintf "error loading configuration, please run 'configure' subcommand (see '%s configure --help' for details)" P.name in
-           Error (`Msg msg))
-      "compile and dynlink"
+         Bos.OS.Cmd.run cmd)
+      "compile configuration"
+
+  (* attempt to dynlink the configuration file.
+   * It is responsible for registering an application via
+   * [register] in order to have an observable
+   * side effect to this command *)
+  let dynlink file =
+    let file = Dynlink.adapt_filename (Fpath.to_string file)
+    and root = get_root ()
+    in
+    try Ok (Dynlink.loadfile Fpath.(to_string (root / "_build" / file)))
+    with Dynlink.Error err ->
+      let err = Dynlink.error_message err in
+      let msg = Printf.sprintf
+          "error %s while loading configuration, please run 'configure' \
+           subcommand (see '%s configure --help' for details)" err P.name
+      in
+      Error (`Msg msg)
 
   module Config' = struct
     let pp_info (f:('a, Format.formatter, unit) format -> 'a) level info =
@@ -598,7 +607,8 @@ module Make (P: S) = struct
   end
 
   let load' file =
-    compile_and_dynlink file >>= fun () ->
+    compile file >>= fun () ->
+    dynlink file >>= fun () ->
     registered () >>= fun t ->
     Log.info (fun m -> m "using configuration %s" (Config.name t));
     Ok t
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -684,59 +684,60 @@ module Make (P: S) = struct
     let full_eval = Cmd.read_full_eval argv in
 
     (*    (c) the config file passed as argument, if any *)
-    (match Cmd.read_config_file argv with
-     | None -> ()
-     | Some cfg -> set_config_file cfg);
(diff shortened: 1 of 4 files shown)


hannesm:
this does not fix https://github.com/mirage/mirage/issues/732 nor https://github.com/mirage/mirage/issues/713, but instead re-enables `mirage help` (and subhelps, such as `mirage help build`) to work without a `config.ml` file around!

hannesm:
I pushed some more commits:

in an empty directory:
```Bash
empty# ls
empty# mirage configure
configuration file config.ml missing
empty# mirage build
configuration file config.ml missing
empty# mirage configure -f foo.ml
Fatal error: exception (Invalid_argument "config must be an existing file (single segment)")
empty# mirage --version
v2.9.0-397-g16c20a2
empty# mirage help | cat
MIRAGE(1)                        Mirage Manual                       MIRAGE(1)



NAME
       mirage - The mirage application builder

SYNOPSIS
       mirage COMMAND ...

DESCRIPTION
       The mirage application builder. It glues together a set of libraries
       and configuration (e.g. network and storage) into a standalone
       unikernel or UNIX binary.

       Use either mirage <command> --help or (mirage) help <command> for more
       information on a specific command.

COMMANDS
       build
           Build a mirage application.

       clean
           Clean the files produced by mirage for a given application.

       configure
           Configure a mirage application.

       describe
           Describe a mirage application.

       help
           Display help about mirage commands.
...

```

a populated unikernel:
```Bash
console# ls
config.ml     unikernel.ml  
console# mirage --version
v2.9.0-397-g16c20a2
console# mirage describe
Name       console
Root       /usr/home/hannes/devel/mirage/mirage-skeleton/console
Keys       logs= (default), target=unix (default), warn_error=false (default)
console# mirage configure
console# mirage build
...ocamlbuild output...
console# mirage clean
console# mirage build
mirage: Configuration is not available. Please run configure

console# echo foo > config.ml
console# mirage configure
error while executing ocamlbuild -use-ocamlfind -classic-display -tags
                        bin_annot -quiet -X _build-ukvm -pkg mirage
                        config.cmxs
+ ocamlfind ocamlc -c -bin-annot -package mirage -o config.cmo config.ml
File "config.ml", line 1, characters 0-3:
Error: Unbound value foo
Command exited with code 2.
```

I guess this solves https://github.com/mirage/mirage/issues/732 and #86 

Additionally, it fixes #72 (`mirage help configure -f tls.ml` works, `mirage configure --help` still complains)

```Bash
console# cp ../static_website_tls/config.ml tls.ml
console# mirage help configure | cat
...
UNIKERNEL PARAMETERS
       -l LEVEL, --logs=LEVEL
           Be more or less verbose. LEVEL must be of the form *:info,foo:debug
           means that that the log threshold is set to info for every log
           sources but the foo which is set to debug.
...
console# mirage help configure -f tls.ml | cat
...
UNIKERNEL PARAMETERS
       --dhcp=DHCP (absent=false)
           Enable dhcp for the unikernel.

       --http=VAL (absent=8080)
           Listening HTTP port.

       --https=VAL (absent=4433)
           Listening HTTPS port.

       --interface=INTERFACE (absent=tap0)
           The network interface listened by the unikernel.

       --ipv4=IPV4 (absent=10.0.0.2/24)
           The network of the unikernel specified as an IP address and
           netmask, e.g. 192.168.0.1/16 .

       --ipv4-gateway=IPV4-GATEWAY (absent=10.0.0.1)
           The gateway of the unikernel.

       --kv_ro=KV_RO (absent=crunch)
           Use a fat, archive, crunch or direct pass-through implementation
           for the unikernel.

       -l LEVEL, --logs=LEVEL
           Be more or less verbose. LEVEL must be of the form *:info,foo:debug
```

hannesm:
imho ready to merge, improves the current state and fixes known bugs.

yomimono:
Thanks for the comment here, you saved me trying all that stuff :P

Drup:
Nice.
I glanced at the code quickly and it looks good.

We should make `mirage configure --help` works, though.

app/functoria_app.ml
```diff
@@ -641,22 +671,22 @@ module Make (P: S) = struct
...
 
-    (*    (c) whether to fully evaluate the graph *)
+    (*    (b) whether to fully evaluate the graph *)
     let full_eval = Cmd.read_full_eval argv in
 
-    (*    (d) the config file passed as argument, if any *)
-    set_config_file (Cmd.read_config_file argv) ;
+    (*    (c) the config file passed as argument, if any *)
+    (match Cmd.read_config_file argv with
+     | None -> ()
```
Drup: Shouldn't this call `handle_parse_no_config`, directly ?

hannesm:
ok, you wanted the `mirage configure --help` story:

once upon a time, [peek_opts](http://erratique.ch/software/cmdliner/doc/Cmdliner.Term.html#VALeval_peek_opts) was called (in `Functoria_command_line.read_config_file`).

```OCaml
val config_file : string Term.t = <abstr>
# Term.eval_peek_opts ~argv:[| "foo" ; "-f" ; "tls.ml" ; "--help" |] config_file;;
- : string option * string Term.result = (None, `Help)
```

which leads in our `read_config_file` to an `invalid_arg`.

We could catch this, and return None.  This would allow further processing, but it would mean that the `get_config_file ()` in `Functoria_app` would return `config.ml`, and not `tls.ml`.

And `get_config_file ()` is used via `load'`, before command line arguments are processed.  But we need to first process the `-f tls.ml`, then `load' tls.ml` in order to show the accurate help text.

Or do the whole `load` business only after the command line parser has done its work.  I tried to implement exactly this for the last hours, but always run into walls.

app/functoria_app.ml
hannesm: did so, doesn't seem to make a difference

Drup:
> Or do the whole load business only after the command line parser has done its work. I tried to implement exactly this for the last hours, but always run into walls.

You can't. You need to evaluate the configuration before knowing how to parse the command line.

hannesm:
There's an easy fix: remove the mention that `mirage configure --help` will do sth useful 7a7135d ;)

And, according to mirage help:
```
       --help[=FMT] (default=pager)
           Show this help in format FMT (pager, plain or groff).
```

Thus, `--help` is just to set the formatter, not to get help.  We only have to educate @mor1 to use `mirage help configure` instead of `mirage configure --help`...

Drup:
I'm not sure if it's a joke or not, so just in case: Thanks, but no.

We should still merge this, it's an improvement, we will just have to revisit the error handling once again to solve that.

hannesm:
it's not a joke.  it's getting way too messy here.  still in favour to remove the `-f` option.

PR #100, opened 2017-01-19
Do not error on `mirage configure --help`

avsm:
This prevents `mirage configure --help` raising an exception
on the CLI. However, it does not parse the config.ml, so the extra
fields only show up if `mirage help configure` is used.  That's not
ideal...

### DIFF
--- a/app/functoria_command_line.ml
+++ b/app/functoria_command_line.ml
@@ -243,6 +243,7 @@ let read_config_file : string array -> Fpath.t option =
         Some (Fpath.v config)
       else
         None
+    | _, `Help -> None
     | _ -> invalid_arg "config must be an existing file (single segment)"
 
 let read_full_eval : string array -> bool option =


amirmc:
Related: https://github.com/mirage/functoria/pull/97#issuecomment-268560723

hannesm:
this seems to solve the problem described.  as noted earlier (https://github.com/mirage/functoria/pull/91, https://github.com/mirage/functoria/pull/97) I'd be in favour to simplify argument processing (by not providing a way to customise `config.ml` to a different filename) (I dislike that `foo --help` and `help foo` vary in their outputs).

avsm:
>  I'd be in favour to simplify argument processing (by not providing a way to customise config.ml to a different filename) (I dislike that foo --help and help foo vary in their outputs).

I'm inclined to support this. It's very confusing indeed to have the two different commands, and the restriction on a single `config.ml` is one that can be lifted in a later release if it proves to onerous.

yomimono:
I believe this is superseded by #101, so closing -- feel free to rebase and reopen if you disagree.

PR #101, opened 2017-01-22
remove possibility to pass `-f <config>` to functoria.

hannesm:
this lets `mirage help configure` and `mirage configure --help` see the same
output (in contrast to the current situation).   It also simplifies argument
processing.

as most recently mentioned in #100, solves #72.  IMHO this improves first user experience a lot, but removes a degree of freedom: each unikernel has to have its configuration in `config.ml`.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -467,10 +467,7 @@ module Make (P: S) = struct
   module Log = (val Logs.src_log src : Logs.LOG)
 
   let configuration = ref None
-  let config_file = ref Fpath.(v "config.ml")
-
-  let get_config_file () = !config_file
-  let set_config_file file = config_file := file
+  let config_file = Fpath.(v "config.ml")
 
   let get_root () =
     match Bos.OS.Dir.current () with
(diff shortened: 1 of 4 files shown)


yomimono:
Thanks for this patch.  As you've mentioned in other issues related to this, the code is indeed simpler if we don't allow users to specify the configuration file, and `mirage configure --help`'s current surprising behavior is indeed a problem which this fixes.

I'm really not a big fan of this solution, though.  I'd like to open up a wider discussion on fixing this on the mailing list or somewhere else (apologies if this already exists somewhere, and if it does please point me to it).  Even if the outcome is "there is no way to support both of these uses without making huge changes to `cmdliner` or creating an entirely new library with a monadic interface", I think it would be useful to come to that outcome more publicly.

It seems that most other people who are paying attention are not as bothered by dropping `-f` as I am, so this patch should probably go in.

avsm:
I'm happy with removing the -f functionality until we know how to put it in properly.  This widens the scope of the tool later, _vs_ baking in confusing functionality today.

amirmc:
It's worth asking/informing via the mailing list before merge.  I believe this question has come up before and it'd be worth checking with any users if they use something other than `config.ml`.  Apologies if I missed such an email already.

PR #104, opened 2017-02-15
Generate smaller, simpler code

yallop:
This PR makes a few changes to the code generated by functoria to make it smaller and simpler.  The aims are readability and simplicity, not performance, but there might be some small performance improvements.

The typical space savings are around 40-45%.  For example,
 * the `main.ml` file for [`conduit_server`][conduit-server] drops from 199 lines to 115 lines
 * the `main.ml` file for [`console`][console] drops from 80 lines to 43 lines

The code is simplified/shrunk in three main ways:

1. First, there's a distinction between between effectful and non-effectful initialization code in the return type of `connect`.  (This is perhaps a first step in the direction of generating code at a slightly higher level than string concatenation.)

   The implementer of `connect` tags the generated code with `Val` or `Eff` according to whether it's a simple value or a possibly-effectful computation, and the caller of `connect` uses the tag to determine whether the value can be safely inlined.  For effectful computations the generated code follows the previous scheme: a top-level binding is emitted
   
   ```ocaml
   let name = lazy (connect_code ())
   ```

   which is used in `connect` code for other modules:

   ```ocaml
   name >>= fun _name ->
   ...
   return (f ... _name ...)
   ```

   For effect-free computations the top-level binding is omitted and the value is instead inlined directly:

   ```ocaml
   ...
   return (f ... connect_value ...)
   ```

2. Second, the code that forces lazy values and the code that executes the resulting `Lwt` computations are combined, avoiding an intermediate binding.  Before:

   ```ocaml   
   let __foo = Lazy.force foo in
   __foo >>= fun _foo ->
   ...
   ```   
   After:
   ```ocaml   
   Lazy.force foo >>= fun _foo ->
   ...
   ```   
   (This changes evaluation order slightly, so it would be a good idea for someone to confirm that the change is safe.)

3. Finally, the layout is now a little more compact.  For example, short expressions now occupy a single line.  Before:

   ```ocaml
   let foo = lazy (
     M.connect ()
    )
   ```
   After:
   ```ocaml
   let foo = lazy ( M.connect ())
   ```

[conduit-server]: https://github.com/mirage/mirage-skeleton/blob/master/conduit_server/config.ml
[console]: https://github.com/mirage/mirage-skeleton/blob/master/console/config.ml

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -42,7 +42,7 @@ let sys_argv = impl @@ object
     method ty = argv
     method name = "argv"
     method module_name = "Sys"
-    method !connect _info _m _ = "return Sys.argv"
+    method !connect _info _m _ = `Eff "return Sys.argv"
   end
 
 let src = Logs.Src.create "functoria" ~doc:"functoria library"
@@ -104,9 +104,10 @@ let keys (argv: argv impl) = impl @@ object
     method !deps = [ abstract argv ]
     method !connect info modname = function
       | [ argv ] ->
-        Fmt.strf
-          "return (Functoria_runtime.with_argv (List.map fst %s.runtime_keys) %S %s)"
-          modname (Info.name info) argv
+         `Eff
+           (Fmt.strf
+              "return (Functoria_runtime.with_argv (List.map fst %s.runtime_keys) %S %s)"
+              modname (Info.name info) argv)
       | _ -> failwith "The keys connect should receive exactly one argument."
   end
 
@@ -141,7 +142,7 @@ let app_info ?(type_modname="Functoria_info")  ?(gen_modname="Info_gen") () =
     val file = Fpath.(v (String.Ascii.lowercase gen_modname) + "ml")
     method module_name = gen_modname
     method !packages = Key.pure [package "functoria-runtime"]
-    method !connect _ modname _ = Fmt.strf "return %s.info" modname
+    method !connect _ modname _ = `Eff (Fmt.strf "return %s.info" modname)
 
     method !clean _i =
       Bos.OS.Path.delete file >>= fun () ->
@@ -163,7 +164,7 @@ let app_info ?(type_modname="Functoria_info")  ?(gen_modname="Info_gen") () =
 
 module Engine = struct
 
-  type tbl_entry = {name: string}
+  type tbl_entry = {name: string; value: string option}
 
   let if_context =
     let open Graph in
@@ -267,10 +268,10 @@ module Engine = struct
     Graph.fold f job @@ R.ok () >>| fun () ->
     tbl
 
-  let meta_init fmt (connect_name, result_name) =
-    Fmt.pf fmt "let _%s =@[@ Lazy.force %s @]in@ " result_name connect_name.name
-
   let emit_connect fmt (iname, (names : tbl_entry list), connect_string) =
+    let meta_init fmt (connect_name, result_name) =
+      Fmt.pf fmt "let _%s =@[@ Lazy.force %s @]in@ " result_name connect_name.name
+    in
     (* We avoid potential collision between double application
        by prefixing with "_". This also avoid warnings. *)
     let rnames = List.map (fun x -> "_"^x.name) names in
@@ -285,7 +286,7 @@ module Engine = struct
       iname
       Fmt.(list ~sep:nop meta_init) (List.combine names rnames)
       Fmt.(list ~sep:nop bind) rnames
-      (connect_string rnames)
+      (let `Eff e = connect_string rnames in e)
 
   let emit_run init main =
     (* "exit 1" is ok in this code, since cmdliner will print help. *)
@@ -306,7 +307,7 @@ module Engine = struct
       | `Impl (c, `Args args, `Deps deps) ->
         let ident = name c (Graph.hash v) in
         let modname = Graph.Tbl.find modtbl v in
-        Graph.Tbl.add tbl v { name = ident };
+        Graph.Tbl.add tbl v { name = ident; value = None };
         let names = List.map (Graph.Tbl.find tbl) (args @ deps) in
         Codegen.append_main "%a"
           emit_connect (ident, names, c#connect info modname)
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -42,7 +42,7 @@ let sys_argv = impl @@ object
     method ty = argv
     method name = "argv"
     method module_name = "Sys"
-    method !connect _info _m _ = `Eff "return Sys.argv"
+    method !connect _info _m _ = `Val "Sys.argv"
   end
 
 let src = Logs.Src.create "functoria" ~doc:"functoria library"
@@ -142,7 +142,7 @@ let app_info ?(type_modname="Functoria_info")  ?(gen_modname="Info_gen") () =
     val file = Fpath.(v (String.Ascii.lowercase gen_modname) + "ml")
     method module_name = gen_modname
     method !packages = Key.pure [package "functoria-runtime"]
-    method !connect _ modname _ = `Eff (Fmt.strf "return %s.info" modname)
+    method !connect _ modname _ = `Val (Fmt.strf "%s.info" modname)
 
     method !clean _i =
       Bos.OS.Path.delete file >>= fun () ->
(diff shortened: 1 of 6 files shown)


yallop:
It may be wise to leave this unmerged until Mirage 3 is safely released.

Drup:
Just so that I can get a quick look before reinstalling half my mirage switch: Could you pastebin before/after for some medium-sized unikernel ?

yallop:
[Here's](https://gist.github.com/yallop/881701355a5182a9c2767bab56d6657d) a before/after gist for [`clock`](https://github.com/mirage/mirage-skeleton/blob/mirage-dev/device-usage/clock/config.ml).

Drup:
Ok, Change 2 is incorrect because it doesn't launch the sub-connect concurently, it makes them sequential.

Here is the original version. First, all the tasks are forced, which make them execute concurrently, then their are lwt-bound.
```ocaml
let f11 = lazy (
  let __time1 = Lazy.force time1 in
  let __pclock1 = Lazy.force pclock1 in
  let __mclock1 = Lazy.force mclock1 in
  __time1 >>= fun _time1 ->
  __pclock1 >>= fun _pclock1 ->
  __mclock1 >>= fun _mclock1 ->
  Unikernel1.start _time1 _pclock1 _mclock1
)
```

Here is the new version, the tasks are forced and lwt-bound immediately, in a sequential order.
```ocaml

let f11 = lazy ( Lazy.force pclock1  >>= fun _pclock1 -> 
 Lazy.force mclock1  >>= fun _mclock1 -> 
Unikernel1.start () _pclock1 _mclock1)
```

Drup:
(To see why this is important more concretely: consider the case where you have two main devices that both launch servers listening on different ports. The "main" Mirage device will connect on both. It must do so concurrently)

yallop:
Thanks for taking a look at this, @drup.

Could you be even more concrete? e.g. with a pointer to some code.

The code that `functoria` currently generates looks quite sequential already, since the Lwt computations are sequenced with `>>=`.  But perhaps there's some additional concurrency introduced when the `lazy` values are forced.

Drup:
@yallop `>>=` does not launch the lwt computation on the left hand side, it only waits for it. The actual promise is launched when we `Lazy.force` it. I'll see if I can come up with a unikernel that exercises this feature.

yallop:
It's ok; I see what you mean.  I'll take another look at combining forcing and binding.

yallop:
Having looked at this a bit more, I think an applicative approach could work well here.

We could generate something like this

```ocaml
pure Unikernel1.start <$> lazy () <$> pclock1 <$> mclock1
```

in place of the current

```ocaml
let f11 = lazy (
  let __time1 = Lazy.force time1 in
  let __pclock1 = Lazy.force pclock1 in
  let __mclock1 = Lazy.force mclock1 in
  __time1 >>= fun _time1 ->
  __pclock1 >>= fun _pclock1 ->
  __mclock1 >>= fun _mclock1 ->
  Unikernel1.start _time1 _pclock1 _mclock1
)
```

With some care it should be possible to arrange to run the applicative so that all the laziness is forced before blocking.

Drup:
@yallop I agree. The code is made to precisely emulate the applicative behavior. :)
The problem is how to make that easy to define at the device level.

Currently, functoria asks for the dsl to provide monad-like operator, but that monad doesn't include the lazyness aspect. I guess we could touch that a bit ?

avsm:
> I think an applicative approach could work well here.

That would be lovely alongside the other code like the cmdliner exposure which has similar idioms.

yallop:
@drup: is there a reason why the ["main" code](https://gist.github.com/yallop/881701355a5182a9c2767bab56d6657d#file-before-ml-L75-L80) forces values sequentially?  Is it important to preserve that behaviour?

talex5:
@yallop  I think that is the list of special init tasks. The sequence seems to be:
- Parse arguments
- Set up logging (depends on the arguments)
- Set up everything else (may use logging)

yallop:
Thanks, @talex5.  Do you think it would be worth making the dependencies explicit in the data flow?  i.e. returning a value from each task that can be used as input to subsequent tasks?

talex5:
I originally made a PR in which init tasks were just normal jobs with dependencies, but it was felt that this list-based system was simpler and good enough.

Drup:
The problem with making that part of the data flow explicit is that suddenly, everyone must depend on the key device. People must remember to declare dependency towards the key device (and, spoiler, they won't). The graph becomes mostly unreadable and you get some other annoying issues, and we don't even really gain anything. 

From an abstract point of view, I would really like it, it would be much more elegant. Currently, there is an ugly hack to find back the key device in the graph and do the init probably. It's a very practical approach. If you have a concrete proposition that doesn't make everything else worse, I'm all hear.

yallop:
Thanks, @drup.  That all makes sense.  Surfacing the dependencies is valuable, but I can see how it could lead to a tangle.  I'll give it some thought.

In the meantime, I'll pull some of the less controversial changes here into a separate pull request.

yallop:
Here's an outline of a possible approach that leaves the device code ignorant of laziness.  The idea is to use an explicit representation of the forced computation to ensure that all forcing of lazy values occurs before all use of `Lwt.(>>=)`.

First, the code for each device is a closed function expression:

```ocaml
(fun x y z ->
 Foo.connect x y z)
```

A function, `lift`, turns these expressions into delayed computations (where `t` is abstract):

```ocaml
val lift : 'a -> 'a t Lazy.t
```

A second function, `<$>`, applies the lifted function to delayed `Lwt` computations:

```ocaml
val ( <$> ) : ('a -> 'b) t Lazy.t -> 'a Lwt.t Lazy.t -> 'b t Lazy.t
```

Finally, `run` runs a fully applied computation:

```ocaml
val run : 'a Lwt.t t Lazy.t -> 'a
```

And here's a simple implementation:

```ocaml
type _ t =
     V : 'a -> 'a t
   | App : ('a -> 'b) t * 'a Lwt.t -> 'b t

let lift v = lazy (V v)

let rec run' : 'b 'c. 'b t -> ('b -> 'c Lwt.t) -> 'c Lwt.t =
  fun e k -> match e with
    | V v -> k v
    | App (f, v) -> run' f (fun g -> v >>= fun w -> k (g w))

let run v = Lwt_main.run (run' (Lazy.force v) (fun x -> x))

let (<$>) f x = lazy (App (Lazy.force f, Lazy.force x))
```

You might then write code like this:

```ocaml
let p = lazy (P.connect ())
let m = lazy (M.connect ())

let d = run (lift connect <$> p <$> m <$> m)
```

The tricky part is ensuring that the thunks are forced before any binding takes place.  Representing the computation using `t`, `V` and `App` takes care of that, since that structure and its evaluation function `run'` know only about `Lwt`, not about `lazy`.  It's not applicative, exactly, but it's applicative-style.

What do you think, @drup?

samoht:
ping @Drup

Drup:
Sorry for the delay! 

@yallop I believe this is a good idea! Your API looks good. I wonder how generic we can make it.
One alternative API would be to have a multi argument apply (with an heterogeneous list) and have an heterogeneous `Lwt.join`. This is pretty much the same anyway (the `<$>` are the cons) but it makes the fact that we want the join semantics a bit clearer.

avsm:
This one seems relevant to resurrect again :-)

Drup:
Indeed! 
@yallop Do you want to take a shot at implementing your last proposition? We could also take that occasion to revisit what we need from the underlying datatype. Currently it's a monad, but maybe we don't need that much.

hannesm:
sorry for being late to the party, but I'm wondering what the status of this PR is? it looks like this would be a good improvement for functoria.

AFAICT @yallop most recent suggestion from https://github.com/mirage/functoria/pull/104#issuecomment-295317334 to use an explicit representation of the forced computation is not yet implemented -- @yallop any chance you have some time to develop this code and rebase this PR on master? I can rebase the accompanied PR in mirage if you like.

Drup:
Actually, the last set of patches implement the idea of separating values and computations. The code looks decent at a glance, but it breaks everything single device in existence. :/

hannesm:
@drup I searched through GitHub for mentions of `base_configurable` (which catches most definitions of custom devices), and most were in mirage/mirage and mirage/mirage-skeleton, a few distributed around other example unikernels and your lua. I think it is manageable to break the API and update all clients.

yallop:
I've rebased against `master`, and I plan to revisit this code to bring it into line with the design discussed above.

hannesm:
thanks @yallop, if you get around doing that, that'd be great. as mentioned earlier, I'm happy to rebase your https://github.com/mirage/mirage/pull/790 PR once this is settled :D

samoht:
It would be nice to merge that PR at one point, as simpler code is always nicer :-)

@yallop do you think you could update these patches to make it inline with the above discussion?

samoht:
Actually this repository is not active anymore, so please feel free to move your patch to mirage/mirage !

PR #108, opened 2017-03-01
Add a --output option to configure

samoht:
That option is used to overwrite the name of the unikernel, eg. this
enables the workflow:

```
mirage configre -o foo
make
./foo
```

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -397,6 +397,7 @@ module Cache : sig
   val clean : Fpath.t -> (unit, [> Rresult.R.msg ]) result
   val get_context : Fpath.t -> context Term.t ->
     [> `Error of bool * string | `Ok of context option ]
+  val get_output: Fpath.t -> [> `Error of bool * string | `Ok of string option ]
   val require :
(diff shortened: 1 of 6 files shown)


samoht:
I am not sure the implementation is the more efficient one,  I've tried to update the info type instead, but I didn't managed to make something cleaner.

/cc @Drup and @yallop

yomimono:
This change looks larger than I'd expect for the feature (which itself is a good idea, IMO).  I'd really appreciate some attention from @drup or @yallop on this PR.

hannesm:
Hmm, out of the blue I'd add the `-o` only for configure, which is then in the cache, and reused for build, depend, and clean!?!

samoht:
I have simplified my patch as suggested by @hannesm

hannesm:
LGTM, thx.  when travis is happy, I think this is good to merge.

Drup:
I gave a very quick look, it looks ok. The additions are a bit sprinkled everywhere, but that's not exactly surprising.

djs55:
Travis is happy, I think we can merge!

PR #115, opened 2017-06-20
Switch to jbuilder

djs55:
This PR proposes to switch from ocamlbuild to jbuilder. Caveats:

- there doesn't seem to be an analog for the `dontlink(unix str num threads)`. The generated `META` file for `functoria-runtime` has `requires = "bytes cmdliner fat result char"` which looks ok. All the `mirage-skeleton` kernels still link ok for `xen`, `unix`, `ukvm` and `virtio` i.e. there's no unexpected linkage to unix / threads etc
- this preserves the existing choice of warnings and errors rather than use the `jbuilder --dev` defaults that we're using elsewhere
- I'm not sure what to do about `doc/index.text`

One nice side-effect is this removes the duplication between the `_tags` file and the `META` file, which had become slightly out-of-sync. The dependencies of each library have been trimmed to only include the direct dependencies.

This PR also includes the `Makefile` boilerplate for preparing multiple package releases.


Drup:
Droplink `dontlink` is fine. It was more a guard against my own stupidity during initial dev than anything else. With the amount of testing we have now, that kind of error will be caught.

As far as I understand, you can replace `index.text` by a `.mld` file with the appropriate odoc option. Not sure how jbuilder handles that though.

What's the diff between the warnings used in functoria and the one proposed by jbuilder's dev mode ?

djs55:
@Drup thanks -- I was hoping you'd say that about `dontlink` :)

To be honest I'm not sure how the docs are handled either -- @samoht: what do you think?

The [jbuilder --dev warnings](https://github.com/janestreet/jbuilder/blob/bdb542ea2970bf9e933a3fe830a5e84c21a88a3a/src/ocaml_flags.ml#L6) seem to be equivalent to `A-4-29-40-41-42-44-45-48-58-59-60` while the current functoria set are `A-40-42-44-48` so I think the `functoria` set are stronger.

Drup:
@djs55 Ok, let's keep warnings that way until https://github.com/janestreet/jbuilder/issues/44 lands, and then we could tweak the dev mode.

samoht:
Currently the `mld` files are used to add custom doc index to individual packages. I think that what we want instead is to use it as the top-level index for the package. Would be useful to add a comment on https://github.com/janestreet/jbuilder/issues/63 to weight in on that use-case :-)

functoria-runtime.opam
```diff
@@ -12,11 +12,16 @@ doc:          "https://mirage.github.io/functoria/"
 license:      "ISC"
 tags:         ["org:mirage"]
 
-build: ["ocaml" "pkg/pkg.ml" "build" "--pkg-name" name "--pinned" "%{pinned}%" ]
+build: [
+  ["jbuilder" "subst" "-p" name "--name" name] {pinned}
```
samoht: This can be simpler now, the man page says to just use `["jbuilder" "subst"] {pinned}`

functoria-runtime.opam
```diff
@@ -12,11 +12,16 @@ doc:          "https://mirage.github.io/functoria/"
...
+]
+build-test: [
+  ["jbuilder" "runtest"]
+]
+
 depends: [
-  "ocamlfind" {build}
-  "ocamlbuild" {build}
-  "topkg"     {build & >= "0.7.3"}
+  "jbuilder" {build & >="1.0+beta9"}
```
samoht: I think `subst` appeared in `beta10`

functoria.opam
```diff
@@ -12,15 +12,15 @@ doc:          "https://mirage.github.io/functoria/"
 license:      "ISC"
 tags:         ["org:mirage"]
 
-build: ["ocaml" "pkg/pkg.ml" "build" "--pkg-name" name "--pinned" "%{pinned}%" "--tests" "false" ]
+build: [
+  ["jbuilder" "subst" "-p" name "--name" name] {pinned}
```
samoht: same here

djs55:
Based on some advice from @avsm on IRC I've added a call to `jbuilder build @doc` and added a Makefile target so `make doc` works. It looks like the only missing piece is the custom doc index, support for which seems to be in-progress over in janestreet/jbuilder#63

avsm:
I'll go ahead and merge this now for an intermediate release in mirage-dev

PR #120, opened 2017-07-04
activate tests in travis ci

avsm:
also do not run tests in the wrong package (functoria-runtime)
as that will try to run the functoria tests but not have
the deps installed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ env:
   global:
   - PACKAGE=functoria
   - PINS="mirage-skeleton.dev:https://github.com/mirage/mirage-skeleton.git"
+  - TESTS=true
   matrix:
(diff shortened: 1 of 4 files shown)


samoht:
Do we still want to test revdeps on every PR to functoria? I think we should add more targeted tests instead.

.travis.yml
```diff
@@ -6,6 +6,7 @@ env:
   global:
   - PACKAGE=functoria
   - PINS="mirage-skeleton.dev:https://github.com/mirage/mirage-skeleton.git"
+  - TESTS=true
```
samoht: actually this doesn't work as we are using a custom TravisCI instead.

samoht:
Let me close this and re-open a new PR using the normal TravisCI scripts without revdeps testing. I'll cherry-pick some of your changes.

Issue #116, opened 2017-07-03
Keys are not properly preserved from "configure" to "build"


hannesm:
As reported by @ricarkol in https://github.com/mirage/mirage/pull/847#issuecomment-312399052

Take a `` `Configure `` phase key, such as `warn_error`: `mirage configure --warn-error true`, and do a `mirage build` afterwards.  The `mirage build` step does not take the `warn_error` as `true` into account at all.  `mirage describe` or `mirage describe --eval` presume `warn_error` to be `false` as well.  The full commandline ends up in `.mirage.config`, but isn't respected/loaded for the `build`.  Additionally, a `let warn_error = true` appears in the generated ~~`main.ml`~~ `key_gen.ml` - my intuition is that this shouldn't be there since it is a `` `Configure ``-only key.

I tried and specified a custom `my_string` key - also using ``phase:`Configure`` - which didn't get transported over to the `mirage build`.

The only `` `Configure `` phase key which is properly transported to `build` seems to be `--target`.

I tried to find the issue within functoria and mirage but without any success.  Someone more familiar with the codebase (@Drup ? @samoht ? ;) may have some ideas what is wrong here.

It is pretty annoying and a showstopper for the `--gdb` support (mirage/mirage#847) @ricarkol implemented.

samoht:
I will have a quick look at it.

Drup:
So you want a configure-time value and use it during the build-time process, is that right ? I believe that used to work, but it's probably not exercised by any of the unikernels in mirage skeleton :(

Configure keys should at least be emitted as constants in the keygen file, so that they can be accessed by the unikernel. Something might have gone wrong when we made the changes for the persistent CLI.

If the problem persists, ping me again after my thesis manuscripts in submitted. :p

hannesm:
@samoht thanks!

@Drup yes, this is a configure-time value which I'd like to use during build-time.

thanks for noting that the emission of configure keys to main.ml is expected.

Drup:
It shouldn't be emitted to `main.ml`, it should be emitted to `keygen.ml` (or whatever the name is now).

PR #121, opened 2017-07-04
Add end-to-end tests

samoht:
My main goal is to be able to run multi-stage steps in the tests. To do so:

- use jbuilder to compiler `config.ml` instead of ocamlbuild
- as we need multi-stage build, `jbuilder` has to see the full content of the tree to be able to link the various stages together. So run the compilation at the root of the tree, and re-introduce a `--config` argument to allow to configure a unikernel living in an other directory
- so far the end-to-end tests just call `... [configure|describe]`, and for a mysterious reason they only the second time they are run. The first time they fail with:
    ```
    error interface mismatch on Test_app while loading configuration, please run 'configure' 
    subcommand (see 'test_app help configure' for details)
    ```
    but the second time the tests are invoked everything works which I don't understand yet.

I really would like to have tests able to detect #116 as it becomes a bit hard to maintain that nice (but complex) library otherwise :-)


samoht:
Closing as I have a simpler and working patch set coming in soon.

PR #122, opened 2017-07-04
Cleanups

samoht:


Drup:
Wait, why did you remove the custom warnings and additional flags ?
There was already a discussion about that in the jbuilderification of functoria.

samoht:
Ha yes, I remembered about this, but wasn't sure. I agree with https://github.com/mirage/functoria/pull/115#issuecomment-309755347 so I'll revert these patches.

PR #123, opened 2017-07-04
Re-add the `--config` argument to select which config file to use

samoht:
blah --config foo/bar.ml` is the same as:

```
cd foo && cp bar.ml config.ml && blah
```

/cc @hannesm who removed the CLI argument at the first place

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -477,8 +477,20 @@ module Make (P: S) = struct
   let src = Logs.Src.create P.name ~doc:"functoria generated"
   module Log = (val Logs.src_log src : Logs.LOG)
 
-  let configuration = ref None
+  (* this needs to be set-up beforce any calls to {!register} *)
+  let root = ref None
   let config_file = ref Fpath.(v "config.ml")
+
+  let init_global_state argv =
+    root := None;
+    config_file := Fpath.(v "config.ml");
+    ignore (Cmdliner.Term.eval_peek_opts ~argv Cmd.setup_log);
+    ignore (Cmdliner.Term.eval_peek_opts ~argv @@
+            Cmd.config_file (fun c -> config_file := c));
+    ignore (Cmdliner.Term.eval_peek_opts ~argv @@
+            Cmd.root (fun r ->
+                let (_:bool) = R.get_ok @@ Bos.OS.Dir.create ~path:true r in
+                root := Some r))
 
   let get_cwd () =
     match Bos.OS.Dir.current () with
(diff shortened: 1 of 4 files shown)


hannesm:
sounds great - apart from the introduction of a `ref` for no obvious good reason.  Is the issue we solved back then (https://github.com/mirage/functoria/pull/101#issue-202417020) fixed (namely that `mirage help configure` and `mirage configure --help` didn't output the same (see #72 and #100 for some more details))?

Drup:
Without looking carefully at the code: In which way is this not broken like all the other attempts we have made at this?

samoht:
> Without looking carefully at the code: In which way is this not broken like all the other attempts we have made at this?

I guess that's because the code has been cleaned up a lot and I am just doing the same thing as for the basic log flags. But that's a fair question and I will check that I've broke nothing by adding some tests :p

samoht:
I have added more commits to:
- rename `--config` into `--file` to get some sort of backward compat
- add a `--root` option to select the directory to build the application. Note that `config.ml` will always be build in the directory where `config.ml` lives, only `main.ml` and other build scripts are affected by `--root`

In #121 I've added some tests which uses both `--root` and `--file` and they seem to work fine. I haven't tested much yet: just that all subcommands work and that `foo configure --help` is mostly the same as `foo help configure`. If you need more tests to be re-assured that this is working fine, please send me some ideas :-)

Regarding @hannesm question about side-effects: unfortunately the root and config filename needs to be known when a dynlinked module calls `register`. We currently have no other ways to pass this information to that function that using global references. That's a bit sad, and the API should maybe evolve a bit (by threading the current context somehow in there) but I am not sure this is worth breaking all the existing unikernels which are using the `register` function in their `config.ml`.

hannesm:
On 05/07/2017 16:14, Thomas Gazagnaire wrote:
> In #121 I've added some tests which uses both `--root` and `--file` and they seem to work fine. I haven't tested much yet: just that all subcommands work and that `foo configure --help` is mostly the same as `foo help configure`. If you need more tests to be re-assured that this is working fine, please send me some ideas :-)

the issue we had earlier that keys were only present in `mirage help
configure`, but not in `mirage configure --help` IIRC.

what does "mostly" mean?

samoht:
The current differences are:
- the `SYNOPSIS` section is different
- the `COMMON OPTIONS` is different: one shows a `--help` option and the other don't

But the output is the same for the "interesting" sections: https://github.com/mirage/functoria/pull/124/commits/2c6577c8226d6a26334cdf31e1ef0eb5dbdff31f#diff-05ec3039a4a0ddff049163ca9abcf814R298, especially the keys.

I'll add more test to check that we have a proper error message when `foo help` is called wen `config.ml` is not found.

samoht:
I am merging this, but I think the main initial cause of brokenness was due to scanning for configuration files which was removed in #91

PR #124, opened 2017-07-04
Add end-to-end tests

samoht:
They all pass but they don't do much currently: they just call the various CLI sub-commands.

I plan to add more complete tests in a subsequent PR, to properly test #116 before proposing a fix.


samoht:
Ok I think I am happy with the current level of tests for configure. @Drup and @hannesm are you ok if I merge this?

I will try to write more tests for `describe` later. Then I can finally try to look at #116 :-)

PR #131, opened 2017-07-05
app: turn the packages and ignored directories used to compile config…

samoht:
….ml into parameters

PR #134, opened 2017-07-06
Rename --root into --build-dir and fix it

samoht:
Also, add passing tests.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -332,11 +332,11 @@ module Config = struct
 
   type t = {
     name     : string;
-    root     : Fpath.t;
-    packages: package list Key.value;
-    keys    : Key.Set.t;
-    init    : job impl list;
-    jobs    : Graph.t;
+    build_dir: Fpath.t;
+    packages : package list Key.value;
+    keys     : Key.Set.t;
+    init     : job impl list;
+    jobs     : Graph.t;
   }
 
   (* In practice, we get all the keys associated to [if] cases, and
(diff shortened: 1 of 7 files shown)


Drup:
Ok, no, seriously. Please do one PR per feature .... not 5. The time I figure out in which order I should review your last 4 PR, you already merged 2 additional one that changes half the stuff you did previously. I can't review that.

samoht:
> The time I figure out in which order I should review your last 4 PR, you already merged 2 additional one that changes half the stuff you did previously.

???

I have created a PR per independent feature, that you can review in any order:
- add err/help channels: #130 (and follow up is #132 where I've missed one case in the previous PR)
- preserve `argv` not `Sys.argv`: #129 
- more fine-grained control over the compilation of `config.ml`: #131 

The "big" PR is #123, which re-adds `--file` and also add `--root` to control the location of the build directory. Both options enables to write some end-to-end tests: which have been merged in #124.

Now that we a test infrastructure in place, I've added more tests and found some issues that this PR is fixing, as a follow-up on #123.

I am not sure what you want me to do differently. Without any tests in place it's pretty impossible to modify functoria, as it's basically impossible to check for regressions. I have merged a few PRs quickly to be able to have a minimal infrastructure for running end-to-end tests (which were not the case before and have caused a lot of pain). For now on, my goal is to add more tests, and fix anything that can be found.

Drup:
I appreciate immensely the fact that we now have proper testing. I would just like to know what you are adding and why. All the changes have one goal: adding end to end tests. Your answer proves that there is an organization, so why didn't you bother making it clearer before ? You could have packaged all that neatly in a single PR that presents everything in a way that makes sense with the short explanation you just gave.

Now I have to jump around in multiple PRs and review something littered with merge commits, which increases the likely hood of me not even bothering.
I'm sorry, I'm not a machine. If you want reviews (and other people to have a slight idea of how functoria works), organize your commit and your PRs in a way that makes sense to other people.

app/functoria_app.ml
```diff
@@ -548,7 +548,29 @@ module Make (P: S) = struct
...
+          match Fpath.basename target with
+          | "_build" -> Ok ()
+          | b        ->
+            if Fpath.basename !config_file = b then Ok ()
+            else (
+              let dst = Fpath.(build_dir / basename target) in
+              Bos.OS.Path.symlink ~force:true ~target dst
+            ))
+        (Ok ())
+    )
```
Drup: Does this also symlink recursively? Do we want to?

Drup:
Presentation remarks aside, This looks ok (although I will have to actually read our new fancy end-to-end tests later on).

app/functoria_app.ml
samoht: `Bos.OS.Path.matches` also match directories so directories will be symlinked. So I don't think there is a need to do recursive symlinks.

samoht:
Thanks for the review!

PR #138, opened 2017-07-06
Fix persistence of non "if" keys between stages

samoht:
This PR adds:

- a failing test case to reproduce #116 
- a possible fix in dc5e402

I don't know why we were only considering the `if` keys in `Config.keys` and not the full set of keys. This was causing the cross-stage persistence merging to fail. This seems to have been added in [213cec](https://github.com/mirage/functoria/commit/4c95c62d5b94adbc5284f73455fdd93b286018fe#diff-213cec9596caee8880ecbae9462829ebR457) without much explanations, but I guess it was to make `describe` show only "useful" keys. Not sure if that's really what we want anyway.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -758,60 +758,60 @@ module Make (P: S) = struct
          3. an attempt is made to access the base keys at this point.
             when they weren't loaded *)
 
-      match load' !config_file with
-      | Error (`Invalid_config_ml err) -> exit_err (Error (`Msg err))
-      | Error (`Msg _ as err) ->
-        handle_parse_args_no_config ?help_ppf ?err_ppf err argv
-      | Ok config ->
+    match load' !config_file with
+    | Error (`Invalid_config_ml err) -> exit_err (Error (`Msg err))
+    | Error (`Msg _ as err) ->
+      handle_parse_args_no_config ?help_ppf ?err_ppf err argv
+    | Ok config ->
 
-        let config_keys = Config.keys config in
-        let context_args =
-          Key.context ~stage:`Configure ~with_required:false config_keys
+      let config_keys = Config.keys config in
+      let context_args =
+        Key.context ~stage:`Configure ~with_required:false config_keys
+      in
+
+      let cached_context = Cache.get_context config.build_dir context_args in
+      let merge_output term =
+        match Cache.get_output config.build_dir with
+        | `Ok (Some o) ->
+          let update_output (r, i) = r, Info.with_output i o in
+          Cmdliner.Term.(app (const update_output) term)
+        | _ -> term
+      in
+
+      let context =
+        match Cmdliner.Term.eval_peek_opts ~argv context_args with
+        | _, `Ok context -> context
+        | _ -> Functoria_key.empty_context
+      in
+
+      (* 3. Parse the command-line and handle the result. *)
+
+      let configure =
+        Config'.eval ~with_required:true ~partial:false context config
+      and describe =
+        let context = Cache.merge ~cache:cached_context context in
+        let partial = match full_eval with
+          | Some true  -> false
+          | Some false -> true
+          | None -> not (Cache.present cached_context)
         in
+        Config'.eval ~with_required:false ~partial context config
+      and build =
+        Config'.eval_cached ~partial:false cached_context config
+        |> merge_output
+      and clean =
+        Config'.eval_cached ~partial:false cached_context config
+        |> merge_output
+      in
 
-        let cached_context = Cache.get_context config.build_dir context_args in
-        let merge_output term =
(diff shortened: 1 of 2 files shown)


Drup:
The notion of primary was present approximately since the beginning (the commit you link only changed where it was used) and it's more about semantics than documentation (although it probably helps).

Basically, we operate in stages: 1) Obtain the values of the if keys with the CLI. 2) Reduce the unikernel description by evaluating all the ifs. 3) Get the resulting non-if keys and evaluate them. This raises errors for missing keys, notably 4) configure.

Basically, Your changes take a mix of all the keys ever declared in your unikernel, including for example, both the keys in the xen and the unix http stack. I decided fairly early on that this was probably not a good idea at all: What happens in term of errors if you have keys that are available in configurations that is not the one currently chosen ? There is also no guarantee that keys in incompatible configurables are not in conflicts.

Additionally ... I don't understand how this can fix our issue :p

samoht:
Ha I see, and that indeed makes sense to not take all the keys.

The issue I tried to fix (and I realise now that it is slightly different from #116) is that the context computed in https://github.com/mirage/functoria/blob/master/app/functoria_app.ml#L772 doesn't include the keys defined in the main job, e.g. https://github.com/mirage/functoria/blob/master/tests/lib/test_app.ml#L48-L51 do not appear in the cached context.

I will try to see if just adding these keys is enough, and I will try to see if I can reproduce with some custom-defined keys.

samoht:
Ok, I've now pushed a much simpler fix.

Drup:
Why is it useful to have those keys resolved during step 1), instead of step 3) like the other keys ?

samoht:
I am not sure to understand how to do this differently. The issue is that `context_args` in https://github.com/mirage/functoria/blob/master/app/functoria_app.ml#L772 doesn't have the project keys defined there. Would you recommend adding these keys in https://github.com/mirage/functoria/blob/master/app/functoria_app.ml#L767 instead of modifying the `Config.key` function? Doesn't seem that anyone else is using that function anyway.

Drup:
So, If I understand correctly, the problem is that the cached context is created between step 1) and step 2).
This is necessary if we want the cache to affect the resolution of the ifs, but it's problematic because we don't have all the other keys yet.

Maybe we should query the cache again after step 3), just like we query the command line several time ?

samoht:
ok so new tentative: b4a1e99 doesn't change the context loaded in 1), it just makes sure that the cached context is loaded with the full set of possible keys. As the cached context has been saved during the configure phase, normally the set of cached keys should be consistent so we will not load some crap.

Also, I noticed that the cache is actually reading and loading the file for every non-if keys. It would probably be a bit wiser to cache this in memory somehow.

Drup:
That might have a chance to work! I still feel a bit insecure about taking all the keys, even if it does sound ok in this case. There still might be a problem with incompatible keys in distinct configurables.

I would feel more comfortable going the other way and just loading things a first time with only the if context, then loading a second time with the full set of keys. This would mimic what we do for the normal CLI.

I don't understand your testcase, are you missing some files ?

samoht:
For the test case: the build step is writing the project keys in some files [here](https://github.com/mirage/functoria/blob/master/tests/lib/test_app.ml#L80), so the tests just read that file and check that they have the right values.

> There still might be a problem with incompatible keys in distinct configurables

I don't see how this could happen: the cached context is always built from a `foo configure` CLI invocation, so would fail if that's invalid.

> I would feel more comfortable going the other way and just loading things a first time with only the if context, then loading a second time with the full set of keys. This would mimic what we do for the normal CLI. What do you think about that solution?

I will have to dig a bit more, but it seemed to me that step 3) was buried deep into the `Config's.eval_cached` call. I will try to understand that code a bit more and try to come-up with something :-)

Drup:
> I don't see how this could happen: the cached context is always built from a foo configure CLI invocation, so would fail if that's invalid.

My worry is not about what appears on the CLI, but the fact that we would build a cmdliner term with overlaping subterms.

samoht:
ok, new fix proposal is in 4826f1 where I try to do what you suggested. And that seems to work, yay! 🎉

Drup:
I think the last commit looks good! The change is rather short and clear in the end.

Is it possible to add a test for keys that are in a conditionally included configurable ?

samoht:
I have started adding tests for custom keys in an other branch (built on top of that one), which is failing for other reasons ... I'll add nested keys too (with conflicting names/types just for fun). Do you want me to add the new tests in this in the PR or can I add them in the PR where I am trying to fix the custom keys?

Drup:
We can just merge this one, yeah.

hannesm:
LGTM, thanks!  haven't tested whether it solves #116, but I believe in the supplied test cases :)

PR #141, opened 2017-07-09
Fix `... help configure` when custom keys are used

samoht:


app/functoria_app.ml
```diff
@@ -812,6 +812,11 @@ module Make (P: S) = struct
       and clean =
         Config'.eval_cached ~partial:false cached_context config
         |> set_output config
+      and help =
+        let context = Cache.merge ~cache:cached_context context in
+        let info = Config.eval ~partial:false context config in
+        let keys = Key.deps info in
+        Key.context ~stage:`Configure ~with_required:false keys
```
samoht: maybe this should somehow be refactored with `Config'.eval_cached`

tests/lib/test_app.ml
```diff
@@ -87,3 +88,7 @@ module C = struct
 end
 
 include Functoria_app.Make(C)
+
+let register ?packages ?keys name jobs =
+  let init = [Functoria_app.(keys sys_argv)] in
+  register ?keys ?packages ~init name jobs
```
samoht: Maybe it would make sense to add `Functoria_app.key` as the default `init` device.

PR #143, opened 2017-08-01
prepare 2.2.0 release

avsm:


avsm:
failure is due to the mirage revdeps not having an upper bound; being fixed in https://github.com/ocaml/opam-repository/pull/9970

PR #150, opened 2018-02-22
Add support for projections

Drup:
This adds a new combinator:

```ocaml
val proj: ('a -> 'b) typ -> string -> ('a -> 'b) impl
```

`proj (foo @-> bar) "Bar"` should be applied to an impl of type `foo` (supporting a module `M`), returns `M.Bar` and the resulting implementation is assumed to be of type `bar`. the connect for this implementation is the identity.

The combinator (and the implementation) is somewhat raw at the moment. It is not very useful for mirage in general, but it made [functoria-lua](https://github.com/Drup/functoria-lua) much cleaner.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -164,14 +164,14 @@ module Engine = struct
     let open Graph in
     Graph.collect (module Key.Set) @@ function
     | If cond      -> Key.deps cond
-    | App | Impl _ -> Key.Set.empty
+    | App | Impl _ | Proj _ -> Key.Set.empty
 
   let keys =
     let open Graph in
     Graph.collect (module Key.Set) @@ function
     | Impl c  -> Key.Set.of_list c#keys
     | If cond -> Key.deps cond
-    | App     -> Key.Set.empty
+    | App | Proj _ -> Key.Set.empty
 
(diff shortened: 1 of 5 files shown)


app/functoria_app.ml
```diff
@@ -197,13 +197,13 @@ module Engine = struct
     Fmt.pf fmt "%s%a"
       c#module_name
       Fmt.(list (parens @@ of_to_string @@ Graph.Tbl.find tbl))
-      args
-
+      args      
+
```
hannesm: i can see trailing whitespaces.

hannesm:
would be nice to have this functionality tested as part of the testsuite, or?

Drup:
@hannesm Absolutely, I won't finish polishing that right away though.

Drup:
On one hand, the mirage set of modules adapted to the absence of this combinator, so it's not needed .... on the other hand, maybe it would avoid akwardness in some modules to have it...

In any case, it's not essential, and there does not seem to be a huge pressing need for it, since nobody has ever requested it.

samoht:
Actually this repository is not active anymore, so please feel free to move your patch to mirage/mirage :-)

PR #151, opened 2018-03-23
expose the transitive opam dependencies of a project

hannesm:
the current state is that the explicitly named opam dependencies are gathered in `Functoria_info`.  While this is useful for building a unikernel, at runtime it is much more convenient to have a complete list of all runtime dependencies (maybe even including build time dependencies) available.  One use case is to be able to poke at a unikernel image and discover whether it needs to be updated (given a new stream of released libraries).  Another use case is to be able to reproduce a unikernel just from a binary (this requires to include `cc`, `ld`, and `as` version information - but not more AFAICT, maybe `CFLAGS` set in the environment).

The approach taken here uses `ocamlfind query -r -format %p` during build, and using `opam subst` afterwards to replace the version numbers with the installed opam packages.

Drawbacks:
- the list of opam repositories is not included (best done including the latest commit/modification -- if anyone has some `opam` command line to get this information, that'd be great -- `opam repo list` is rather incomplete and not too useful)
- the list generated by `ocamlfind` is incomplete (since it uses `ocamlfind` dependencies, as specified in `META`, which is incomplete -- e.g. there's no mention of `ocaml-freestanding` in `mirage-solo5`, `bigarray` requires `unix`, `gmp-xen`/`gmp-freestanding` don't appear in any dependency chain, and I can't find any `ocaml-src` dependencies anywhere)
- for reproducibility, build dependencies may also be crucial: a `ppx` at version X may generate different output than at version Y.  I'd prefer to separate the build and runtime dependencies though (whereas `opam list --recursive --resolve=tcpip` doesn't distinguish between the two).

Previous discussion at https://github.com/mirage/mirage/issues/896 where (a) @avsm mentions the need for the opam repositories, (b) @samoht points to https://github.com/MagnusS/mirage-stats-demo/blob/master/manifest/manifest.ml for earlier work (using `opam list -s --required-by %s --rec --depopts --installed` which includes build dependencies AFAICT), (c) @Drup suggests to use opam2: `opam list --nobuild --depopts --installed --rec --required-by <PKG>` - but I don't think that relying on opam2 is an option atm.

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -115,8 +115,6 @@ let keys (argv: argv impl) = impl @@ object
 type info = Info
 let info = Type Info
 
-(* hannes is pretty sure the following pp needs adjustments, but unclear to me
-   what exactly to do here? i.e. who reads the formatted stuff in again? *)
 let pp_libraries fmt l =
   Fmt.pf fmt "[@ %a]"
     Fmt.(iter ~sep:(unit ";@ ") List.iter @@ fmt "%S") l
@@ -127,13 +125,13 @@ let pp_packages fmt l =
          (fun fmt x -> pf fmt "%S, \"%%{%s:version}%%\"" x x)
         ) l
 
-let pp_dump_info module_name fmt i =
+let pp_dump_pkgs module_name fmt (name, pkg, libs) =
   Fmt.pf fmt
     "%s.{@ name = %S;@ \
      @[<v 2>packages = %a@]@ ;@ @[<v 2>libraries = %a@]@ }"
-    module_name (Info.name i)
-    pp_packages (Info.package_names i)
-    pp_libraries (Info.libraries i)
+    module_name name
+    pp_packages (List.sort String.compare (String.Set.elements pkg))
+    pp_libraries (List.sort String.compare (String.Set.elements libs))
 
 let app_info ?(type_modname="Functoria_info")  ?(gen_modname="Info_gen") () =
   impl @@ object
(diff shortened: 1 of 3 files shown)


Drup:
You seem to be inferring the opam package from the ocamlfind dependencies. That's completely wrong, you shouldn't do that. Infer the opam dependencies using opam! ocamlfind and opam packages are only
very loosely connected in today's ecosystem. It's best to treat them as separate entity and use the appropriate tools to take the transitive closure for each.

hannesm:
@Drup thanks for your feedback. I find your choice of words insulting ("completely wrong", "you shouldn't do that") -- I'm open for constructive feedback, but not for insults.

As the build process of MirageOS goes: it calls `ocamlbuild -use-ocamlfind` (see line 1817 of mirage.ml atm, receiving a list of `libs` (which is `Info.libraries` -- i.e. the ocamlfind libraries), and cflags passed via `pkg-config`) -- opam packages and their dependencies are not used for the compilation and final linking!  If we would use opam , that information would be slightly wrong, though likely an overapproximation (as long as you always `mirage build` in a fresh switch to avoid other packages being required by ocamlfind, but not opam).

On the other hand a user looks at opam packages (including their version numbers and dependencies), and is interested in these releases, rather than ocamlfind.  This is why we end up having to map ocamlfind libraries and version numbers to opam packages, and this PR is an attempt at that.

"use the appropriate tools" well, please educate me what are the appropriate ones.  As mentioned above, I don't think requiring opam2 is an option for now.

We should keep in mind what is the current state, what is the goal, and how to get there!?  I don't think the current state (released functoria) is of much use for the cases outlined above (reproducibility, figuring out whether a recompilation is required), and this PR improves this information, but does not yet reach the goal.
As mentioned above, the reproducibility clearly needs more information (system tools), and the drawbacks of the ocamlfind vs opam inconsistencies are described as well - I'd try to get rid of the inconsistencies between ocamlfind and opam by improving the `META` information of packages (and/or get rid of ocamlfind at a later stage) - which may fail due to our "cross-compilation" approach atm.

Other inconsistencies I just noticed include `tcpip` now `requires: tcpip.unix` (why? this didn't use to be the case in earlier releases, see https://github.com/mirage/mirage-tcpip/blob/v3.1.4/lib/META).

Drup:
I'm sorry if you feel insulted, that was not my intention, but I don't feel like I have been particularly mean or anything.

But trying to map ocamlfind to opam packages by *name* is doomed to fail. The current ecosystem is far too irregular for that. You might aim to make the ecosystem more regular, and I would agree, but that's a long term battle, not something you can rely on now.

The command line I gave you is for opam2 simply because that's what I have installed. :)
The command line used by `mirage-stat-demo` is pretty much the exact same one, but without `--nobuild`. Is it so bad to include build dependencies for now ?

@AltGr Do you have a proposition to fetch all the *installed* non-build recursive dependencies of some currently-installed packages that is compatible with 1.2 ? With depopts ?

AltGr:
> fetch all the installed non-build recursive dependencies of some currently-installed packages

not _per se_, but `opam-lock` or `opam-bundle` do parts of this.
Also, `opam list -s --owns-file $(opam var prefix)/lib/LIBNAME` might help ?

And there is the obvious: `opam list --installed --nobuild --rec --required-by PKG` I guess ?

> that is compatible with 1.2

No.

Drup:
Yeah, I proposed the "obvious" above. `--owns` is also 2.0 only, right ?

> No.

:disappointed: 

Second attempt: we can fetch all the deps in 1.2 easily. Would it be possible after the fact to know which one are build (using some script, if necessary) ?

AltGr:
opam 1.2 already has json output, although less complete, so it could probably extracted from that, yes.

app/functoria_app.ml
```diff
@@ -127,13 +125,13 @@ let pp_packages fmt l =
...
+let pp_dump_pkgs module_name fmt (name, pkg, libs) =
   Fmt.pf fmt
     "%s.{@ name = %S;@ \
      @[<v 2>packages = %a@]@ ;@ @[<v 2>libraries = %a@]@ }"
-    module_name (Info.name i)
-    pp_packages (Info.package_names i)
-    pp_libraries (Info.libraries i)
+    module_name name
+    pp_packages (List.sort String.compare (String.Set.elements pkg))
+    pp_libraries (List.sort String.compare (String.Set.elements libs))
```
Drup: According to the doc, `Set.elements` is guaranteed to be sorted. You could either remove the sorting, or remove the transformation to a set.

hannesm: fixed in 520079e (removed `List.sort String.compare`)

PR #152, opened 2018-05-18
Fix support for negative ints in keys

samoht:
/cc @Drup

### DIFF
--- a/lib/functoria_key.ml
+++ b/lib/functoria_key.ml
@@ -51,7 +51,7 @@ module Arg = struct
 
   let int = conv
       ~conv:Cmdliner.Arg.int ~runtime_conv:"Cmdliner.Arg.int"
-      ~serialize:(fun fmt -> Format.fprintf fmt "%i")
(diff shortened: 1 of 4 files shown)


samoht:
(Otherwise `key_gen.ml` will generate code like `foo -1`)

lib/functoria_key.ml
```diff
@@ -51,7 +51,9 @@ module Arg = struct
 
   let int = conv
       ~conv:Cmdliner.Arg.int ~runtime_conv:"Cmdliner.Arg.int"
-      ~serialize:(fun fmt -> Format.fprintf fmt "%i")
+      ~serialize:(fun fmt i ->
+          if i > 0 then Format.fprintf fmt "%i" i
+          else Format.fprintf fmt "(%i)" i)
```
Drup: Just always put parens everywhere. :)

Drup:
I can't believe it took us 3 years to trigger that. I blame the fact that we are not using a proper metaprogramming tool!

It's clearly correct, but I prefer the more regular version where we always put parens.

PR #157, opened 2018-09-22
output opam2 file

hannesm:
now that opam2 is the default, is there any reason to output a opam 1.2 file!?

### DIFF
--- a/lib/functoria.ml
+++ b/lib/functoria.ml
@@ -184,7 +184,7 @@ module Info = struct
 
   let opam ?name ppf t =
     let name = match name with None -> t.name | Some x -> x in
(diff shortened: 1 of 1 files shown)


Drup:
Do we really want to merge that already ? Not wait a few more weeks until people have migrated ?

hannesm:
Dear @Drup, IIUC the opam-repository won't be updated for opam 1.2.x (i.e. the 1.2 branch will stay there, but won't be updated). And a Mirage 3.2.0 release is coming soon.  Since this will only be available for opam 2.0 users, why should we wait "a few more weeks" for switching to opam2?

samoht:
Do the `depends` field also add the right OCaml versions?

PR #158, opened 2018-10-04
Convert from jbuilder to dune

emillon:
This uses `dune`-the-binary, `dune`-the-file-format as well as some idioms (`(test)` stanza, `:standard`). I'm not sure if it's worth passing `-safe-string` by hand these days - if that's not the case we can just remove `(flags)` completely.


dune-project
```diff
@@ -0,0 +1 @@
+(lang dune 1.0)
```
samoht: missing `(name functoria)`

emillon: good catch, fixed!

hannesm:
@emillon I'm fine with using `dune` defaults, and not pass `-safe-string` manually in the mirage ecosystem

emillon:
@hannesm thanks, I removed `(flags)` to get the default ones.

Drup:
Can you remind me of which warning/errors are going to be enable/disable because of this? The warnings were carefully chosen.

emillon:
The warnings numbers correspond to the following:

```
 40 Constructor or label name used out of scope.
 42 Disambiguated constructor or label name (compatibility warning).
 44 Open statement shadows an already defined identifier.
 48 Implicit elimination of optional arguments.
 15 Private method made public implicitly.
```

I'm not sure why `-warn-error -15` is used since by default it is not a fatal warning.

Drup:
No, I meant, what is the difference between our warnings, and the one enable/disabled by dune by default.

emillon:
It seems that dune suppresses the additional following warnings:

    4 Fragile pattern matching: matching that will remain complete even
    if additional constructors are added to one of the variant types
    matched.
    29 Unescaped end-of-line in a string constant (non-portable code).
    41 Ambiguous constructor or label name.
    45 Open statement shadows an already defined label or constructor.
    58 Missing cmx file
    60 Unused module declaration

emillon:
Thanks. I just rebased this on top of master.

PR #159, opened 2018-10-04
Inject global arguments into generated_header

emillon:
This has two advantages: first, code that uses this function can now be tested. This could also be extended so that it supports SOURCE_DATE_EPOCH for reproducible builds.


app/functoria_app.mli
```diff
@@ -145,7 +145,7 @@ end
 (** Code generation helpers. *)
 module Codegen: sig
 
-  val generated_header: unit -> string
+  val generated_header: ?argv:string array -> ?time:float -> unit -> string
```
hannesm: FWIW, I'd prefer Ptime.t here.

PR #160, opened 2018-10-06
Use Ptime for time argument

emillon:
Cf https://github.com/mirage/functoria/pull/159#discussion_r222954358

cc @hannesm 

Is the extra dependency OK?

Thanks!


Drup:
I'm sorry, but do we gain *anything* by that? We don't use Ptime here, we simply embeded the current time in a file, we don't manipulate it, and we certainly don't use a clock, posix or not.

I know we have a sort of moral obligation of using all the good modern minilibs in the ecosystem, but I would prefer if we had an actual reason to do that.

emillon:
If I see `time:float`, I'll have several things to guess: does this expect seconds? microseconds? nanoseconds? what is the epoch? With `Ptime.t` this is documented by the type itself.

samoht:
LGTM

Drup:
> an actual reason for me would be to use Ptime.pp_rfc3339 instead of the [10 lines printing a timestamp](https://github.com/mirage/functoria/blob/9b0fb4eff737491c88cb5424414088f0dc8409f4/lib/functoria_misc.ml#L80-L89).

Ah, that's a bit more convincing to me. :)

lib/functoria_misc.ml
```diff
@@ -75,21 +75,17 @@ module Codegen = struct
   let generated_header ?(argv=Sys.argv) ?time () =
     let t =
       match time with
-      | None -> Unix.gettimeofday ()
       | Some t -> t
+      | None ->
+          begin
+            match Ptime.of_float_s (Unix.gettimeofday ()) with
+            | Some t -> t
+            | None -> failwith "cannot convert current time"
```
hannesm: Can we call directly `Ptime_clock.now ()` here, please.

emillon:
I used `Ptime_clock` - it looks way better than before.

Drup:
I like the current version much more than the original one, since we actually use `ptime` to remove a bunch of code previously needed!

lib/jbuild
```diff
@@ -1,7 +1,7 @@
 (library
  ((name functoria)
   (public_name functoria)
-  (libraries (unix cmdliner rresult fmt astring fpath))
+  (libraries (unix cmdliner rresult fmt astring fpath ptime.clock.os))
```
hannesm: not sure about best current practises: should `ptime` be included here as well, or is transitivity (via `ptime.clock.os`) sufficient?  I'm in favour of the former.

emillon: in general I agree that transitive dependencies are a potential source of bugs, but within a single ocamlfind lib it's probably acceptable.

PR #161, opened 2018-10-06
Add Functoria_key.add_to_context

emillon:
The `context` type is a bit difficult to manipulate in tests because the only way to build it is through `Cmdliner`.

This adds an `add` function which can extend an existing context. Together with `empty_context`, it makes it possible to build arbitrary contexts.


samoht:
LGTM

PR #166, opened 2018-12-22
code generation: fewer bindings, do not evaluate configurables multiple times, run initialization once

hannesm:
When I look into a (mirage-)generated main.ml, I find the following code:
```OCaml
let mirage1 = lazy (
  let __noop1 = Lazy.force noop1 in
  let __noop1 = Lazy.force noop1 in
  let __key1 = Lazy.force key1 in
  let __mirage_logs1 = Lazy.force mirage_logs1 in
  let __f11 = Lazy.force f11 in
  __noop1 >>= fun _noop1 ->
  __noop1 >>= fun _noop1 ->
  __key1 >>= fun _key1 ->
  __mirage_logs1 >>= fun _mirage_logs1 ->
  __f11 >>= fun _f11 ->
  Lwt.return_unit
  )

let () =
  let t =
  Lazy.force noop1 >>= fun _ ->
    Lazy.force noop1 >>= fun _ ->
    Lazy.force key1 >>= fun _ ->
    Lazy.force mirage_logs1 >>= fun _ ->
    Lazy.force mirage1
  in run t
```

Which IMHO has two issues fixed in separate commits in this PR:
- all initialization jobs are evaluated twice, once in `let ()` (bound to `t`), once in `mirage1` -- the reason is that Mirage's `P` (where `P.create` is called by Functoria) registers the initialization jobs as dependencies -- this is _required_, otherwise they are not part of functoria's graph, and it complains during configure step (if you can think of another way to handle this, please let me know -- I'd be more in favour of not having to pass `init @ jobs` to `P.create`, but instead only `jobs`, and retain the `Lazy.force xxx >>= fun _` in `let ()`)
- each configurable is first let-bound to `__XX`, then Lwt-evaluated `__XX >>= fun _X`, and then potentially applied to connect, using `_X`. This is not really intuitive, since now variables which start with an underscore are passed to a function. Furthermore, if depending on the same job multiple times (i.e. `noop1` above), these are unneccessarily forced twice and Lwt-evaluated twice.

This patch generates instead the following code:
```OCaml
[@@@ocaml.warning "-27"]

let mirage1 = lazy (
  Lazy.force noop1 >>= fun noop1' ->
  Lazy.force key1 >>= fun key1' ->
  Lazy.force mirage_logs1 >>= fun mirage_logs1' ->
  Lazy.force f11 >>= fun f11' ->
  Lwt.return_unit
  )

let () =
  run (Lazy.force mirage1)
```

Please let me know what you think (esp. @Drup, but obviously other input welcome) /cc @linse

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -236,7 +236,7 @@ module Engine = struct
     | [x] -> x
     | _   -> invalid_arg "Functoria.find_device: too many devices."
 
-  let build info (_init, job) =
+  let build info job =
     let f v = match Graph.explode job v with
       | `App _ | `If _ -> R.ok ()
       | `Impl (c, _, _) -> c#build info
@@ -244,7 +244,7 @@ module Engine = struct
     let f v res = res >>= fun () -> f v in
     Graph.fold f job @@ R.ok ()
 
-  let configure info (_init, job) =
+  let configure info job =
     let tbl = Graph.Tbl.create 17 in
     let f v = match Graph.explode job v with
       | `App _ | `If _ -> assert false
@@ -291,7 +291,7 @@ module Engine = struct
        run (Lazy.force %s)@]"
       main
 
-  let connect modtbl info (_init, job) =
+  let connect modtbl info job =
     let tbl = Graph.Tbl.create 17 in
     let f v = match Graph.explode job v with
       | `App _ | `If _ -> assert false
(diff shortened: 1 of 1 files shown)


Drup:
Nothing is "evaluated twice". `>>=` doesn't evaluate its left hand side, it simply waits for it. If if was already evaluated, then nothing further happens.

As for the rest, your patch is incorrect, in that it doesn't execute the dependencies concurrently. Please read this thread: https://github.com/mirage/functoria/pull/104

PR #167, opened 2019-02-20
Use dune to build config.ml

TheLortex:
The change is the following: to build `config.cmxs` functoria now symlinks `config.ml` in the `_build` directory, then it generates a `dune` configuration file to be able to build `_build/_build/default/config.cmxs`. It can also grab a `dune.inc` file in order to add options to the configuration build (necessary for tests). 

Tests have been updated to support the modification. They all pass, and I also tested with some Mirage samples. It might be useful to test it on other setups to be sure it doesn't break anything !

(in the mirage/mirage#969 context)


TheLortex:
By the way this enables the possibility for `config.ml` to depend on external libraries but I haven't implemented that yet.

app/functoria_app.ml
```diff
@@ -622,36 +622,53 @@ module Make (P: S) = struct
...
+          let dune_file = if has_dune_inc then ("(include dune.inc)"^dune_file) else dune_file in
+          let write_dune_file = Bos.OS.File.write config_file dune_file in
+          let write_dune_workspace_file = Bos.OS.File.write Fpath.(v "_build" / "dune-project") "(lang dune 1.7)"
+          in
+          (* Build config.cmxa with dune *)
+          let target_file = Fpath.(v "_build" / "default" / file) |> Fpath.to_string in
+          let cmd =
+            Bos.Cmd.(v "dune" % "build" % "--no-print-directory" % "--root" % "_build" % target_file)
          in
+         write_dune_workspace_file >>= fun _ -> write_dune_file >>= fun _ ->
```
hannesm: not sure what to do with potential errors here -- should maybe the files (`dune-project` / `dune`) be removed if they exist before? (the scenario i have in mind is that i do a mirage configure, modify config.ml (adjusting dependencies) and run it again...)

TheLortex: Yes, I fixed that by removing the files if they exist before.

Drup:
I'll take a look at this when I have time. From a quick look, there are various hacks which I'm not fond of, I wouldn't mind input from the usual dune expect, @rgrinberg and @diml.

diml:
Just a few comments after having a quick look at the diff:

- I can see a few references to `_build`, which means that things might not work as expected when the build directory is set to something else. esy does this for instance
- it is usually not recommended to manually pass include directories. Indeed, the compiler might read files in these directories but dune won't know about it, which means that it can break incremental compilation
- paths such as `lib/.test_app.objs/byte` are internal details of the build rules that are not documented. In particular, they are not covered by the semantic versioning of dune and we already changed them twice to accommodate various features, so I don't recommend making assumption about these paths

In general, if you are trying to do something that can't be done easily with the current feature set of dune, the best is to open a ticket on github.com/dune describing what you are trying to do so that we can discuss and see what can be done.

TheLortex:
I agree with you on this. I translated to dune what was previously done with ocamlbuild to dune so these hacks are not of my creation. 

The goal is to build from a `config.ml` the `config.cmxs` file, and then dynlink it with functoria. The way I do it is the following:
* A `_build` directory is created, inside there is a generated `dune` file and `config.ml` is symlinked.
* `dune build --root _build _build/default/config.cmxs` is executed to generate the dynlinkable config.
* Therefore the dynlinkable file is `_build/_build/default/config.cmxs`.

The problem is that during tests, functoria has to launch dune inside a build context, as `config.ml` resides in `_build/default/tests/app`. In order to build `config.cmxs` I need to define a build context inside the test build context, which is therefore a bit hacky in order to get things right: the drawback is that inside the inner build context there's no access to outer libraries such as `test_app`, which forces me to use the include hack.

How could we reduce the complexity in that ?

rgrinberg:
The above sounds like a cross compilation problem to me. You want to generate the config.cmxs in the host context, and then use it to build something in the target context.

diml:
Would it be easier to define the tests if we consider that functoria has been previously installed? If yes, then the `(package functoria)` dependency might be the answer: it gives you an environment in which it looks like functoria has been installed. In particular, you can call dune inside this environment. That's how we write all our integration tests in dune. Combined with cram testing, it's really convenient.

TheLortex:
Thank you for your answers. I've been able to get rid of most hacks using the `(package functoria)` dependency at the cost of defining a `functoria.test_app` package. 
Now the behavior is the following:
- `app/config.ml` and `app/dune.inc` are symlinked in a `build-config/` directory.
- `_build/default/build-config/config.cmxs` is built using dune. In the test case, it's has access to `functoria` and `test_app`. 
- dynlinking is done as before. 

I'll push the update soon.

app/functoria_app.ml
```diff
@@ -622,36 +624,64 @@ module Make (P: S) = struct
...
-           | dirs  -> Bos.Cmd.(v "-Xs" % String.concat ~sep:"," dirs)
-         in
-         let cmd =
-           Bos.Cmd.(v "ocamlbuild" % "-use-ocamlfind" % "-classic-display" %
-                    "-tags" % "bin_annot" % "-quiet" %%
-                    ignore_dirs %% pkgs % file)
+          Log.info (fun m -> m "%s" file_ml);
+          let target_dir = Fpath.(v "build-config")
+          (* Import files in the build target *)
+          and target_path = Fpath.(v "build-config" / file_ml)
```
emillon: what do you think about reusing the `target_dir` variable?

app/functoria_app.ml
```diff
@@ -622,36 +624,64 @@ module Make (P: S) = struct
...
+              in
+              let pkgs = if has_dune_inc then "config_custom"::pkgs else pkgs
+              in
+              String.concat ~sep:" " pkgs
+          in
+          let dune_content_default = "(library (name config) (modules config) (libraries "^pkgs^"))" in
+          let dune_content = if has_dune_inc then ("(include config.dune)"^dune_content_default) else dune_content_default in
+          let write_dune_file = Bos.OS.File.delete dune_file >>= fun () -> Bos.OS.File.write dune_file dune_content
+          and write_dune_project_file = Bos.OS.File.exists dune_project_file 
+          >>= function | false -> Bos.OS.File.write dune_project_file "(lang dune 1.7)"
```
emillon: 1.0 would be a more conservative choice if you don't need fancy features.

tests/test_full.ml
```diff
@@ -82,26 +83,26 @@ let test_configure () =
...
-    ["app.ml"; "config.ml"; "myocamlbuild.ml"; "_build"]
+  Alcotest.(check files) "nothing should be created in the source dir"
+    ["app.ml"; "config.ml"; "config.dune"; ]
     (list_files "app");
   Alcotest.(check files) "other files should be created in custom_build_"
-    ["main.ml"; "app.ml"; ".mirage.config"; "dune"; "key_gen.ml";
-     "myocamlbuild.ml" (* FIXME: add a .mirage-ignore file to avoid this *) ]
+    ["main.ml"; "app.ml"; "config.dune"; ".mirage.config"; "dune"; "key_gen.ml";
+    "build-config"; "dune-project"; "_build"
+   (* FIXME: add a .mirage-ignore file to avoid this *) ]
```
emillon: As a reminder, don't forget this :)

app/functoria_app.ml
TheLortex: No problem with that, so I changed this in last commit.

app/functoria_app.ml
TheLortex: Yes, I updated that !

Drup:
Hmm, is using symlinks a potential issue for portability ?

TheLortex:
@Drup On Windows `Unix.symlink` is a privileged operation (according to OCaml's `Unix` module documentation) so that might be an issue. However there is already some code that uses `Bos.OS.Path.symlink` (https://github.com/mirage/functoria/blob/master/app/functoria_app.ml#L584) therefore we might ignore that for now. I don't see any other portability issue. What do you think ?

avsm:
This looks good enough to merge and work on in `master` for a while...

samoht:
Just a few remarks:
- could you use a `copy` rule in the `dune` file instead of symlinks?
- is `config.dune` mandatory? Can you describe how to use that file to add external dependencies? I think the current norm is to name these files `dune.<smting>` so `dune.config` would be better.

Once this is cleared out, I think we should merge this PR to unlock the dune+mirage port.

rgrinberg:
Let me just note that using the `symlink` action in dune does not cause portability issues. Dune will in fact copy the file anyway on windows.

TheLortex:
@samoht 
>     * could you use a `copy` rule in the `dune` file instead of symlinks?

I encounter some problems with that because `dune.config` and `config.ml` can live outside the dune workspace when the `-build-dir` is option used.

>     * is `config.dune` mandatory? Can you describe how to use that file to add external dependencies? I think the current norm is to name these files `dune.<smting>` so `dune.config` would be better.

Let's use `dune.config` then. This file is optional, and if it exists it's included in the generated dune file, and a dependency to a dummy `config_custom` library is added. To add external dependencies, you just have to specify them in the dummy library specification, inside `dune.config`. This is done for tests, for example:
```
(library
    (name config_custom)
    (libraries functoria.test)
    (modules)
)
```

samoht:
Looks good, thanks!

PR #170, opened 2019-05-09
functoria_app: app_info fix callout to opam list --installed

hannesm:
when I initially wrote the code, the logic was slightly different. then I tried to rebase on opam2 and didn't properly test the code it seems.

the result with functoria 2.2.3 is that `rdeps` is always empty since `opam list ...` expects comma-separated package names (rather than a list of packages). it is also the case that `opam list ...` returns the given packages --> no need to union `packages` with `rdeps`. and furthermore, `opam` likes packages, not ocamlfind-libraries+sublibraries.

TL;DR: this fixes `app_info` to output the whole set (including depopts and build only dependencies) of used packages, which is an overapproximation (but IMHO better than an underapproximation). I've tested this code with several unikernels, it works nicely for me.

there's some struggle with functoria, since it current master has #167 merged which is not yet ready for being released AFAIU, this patch is against 2.2.3 (I'd hope once we dune build, we'll be able to generate a more precise manifest of used packages, and don't need to call out to opam twice (`list` and `subst`)). my suggestion is, if people are happy with this PR, to create a 2.2 branch, push this change on that, and release functoria 2.2.4.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## v2.2.4 (2019-05-27)
+
+* fix app_info - executing "opam list --installed" (#170, by @hannesm)
+
 ## 2.2.3 (21/11/2018)
 
 * fix support for pin-depends (#165, by @hannesm)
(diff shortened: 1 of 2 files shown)


hannesm:
anyone has an opinion on this? is a release of master planned soon? if not, I'd go ahead, create a 2.2 branch from the 2.2.3 release, merge this there and release 2.2.4.

//cc @Drup @samoht

hannesm:
done as described above.

PR #171, opened 2019-05-10
Replace dynlink method by a 2-stage build

TheLortex:
Hi, this is quite a heavy PR. I can try to minimize the diff if you want. 

# Why

The dynlink method to load `config.ml` feels quite fragile, as it requires `mirage` and `config.ml` to be built with the same code / options. When compiling using `dune`, these assumptions can be broken:
- `config.ml` can be built with a local instance of the mirage source code. 
- Link-time optimizations might not be enabled for `mirage` but are enabled for `config.ml`. 

# Proposal

Instead of dynlinking `config.ml`, we can perform a 2-stage build of the configuration !
(1) Locally build `config.ml`
(2) Execute `config.ml`, which calls some entry function in `Functoria_app`

There are several ways of doing it:
* Breaking changes for both `config.ml` and `mirage`:
  - Separate `Functoria_app.Make` in two functors, one that builds the `config.ml` and one that is called by the built `config.ml`
  - Change `register ...` to a more appropriate name 
* Breaking changes for `mirage`:
  - Only perform the separation of `Make` and have `register ...` as the entry point of (2)
* No breaking changes:
  - Keep everything under `Make` and hack with the names that are currently used.

I've implemented the intermediate solution. This is probably far from perfect but I'd like to have feedback on that :)

### DIFF
--- a/app/functoria_app.ml
+++ b/app/functoria_app.ml
@@ -380,7 +380,6 @@ module Config = struct
   let extract_keys impl =
     Engine.keys @@ Graph.create impl
 
-  let name t = t.name
   let keys t = t.keys
 
   let gen_pp pp fmt jobs =
@@ -415,7 +414,7 @@ end = struct
   let save ~argv root =
     let file = filename root in
     Log.info (fun m -> m "Preserving arguments");
-    let args = Array.to_list argv in
+    let args = List.tl (Array.to_list argv) in (* Only keep args *)
     let args = List.map String.Ascii.escape args in
     let args = String.concat ~sep:"\n" args in
     Bos.OS.File.write file args
@@ -486,13 +485,12 @@ module type DSL = module type of struct include Functoria end
 
 module Make (P: S) = struct
 
-  let src = Logs.Src.create P.name ~doc:"functoria generated"
-  module Log = (val Logs.src_log src : Logs.LOG)
+  (* GLOBAL STATE *)
 
   (* this needs to be set-up beforce any calls to {!register} *)
   let build_dir = ref None
+  let default_init = [keys sys_argv]
   let config_file = ref Fpath.(v "config.ml")
-  let build_config_dir = Fpath.(v "build-config")
 
   let init_global_state argv =
     build_dir := None;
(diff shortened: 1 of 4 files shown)


cfcs:
@TheLortex I'm super excited about this since I think it can pave the way for external device types/implementations. I cannot get this to build though, can you link to the corresponding `mirage` branch?

TheLortex:
The PR currently breaks API with Mirage, but it's not necessary, I'm waiting for comments on this.

There are two branches of https://github.com/TheLortex/mirage that implement the change of this PR:
- `functoria-fix` (https://github.com/TheLortex/mirage/tree/functoria-fix):
  - builds unikernels using ocamlbuild, should be stable to test this PR
- `full-dune+functoria-fix` (https://github.com/TheLortex/mirage/tree/full-dune+functoria-fix):
  - new shiny dune build relying on dune 1.10 so that's mostly experimental

app/functoria_app.ml
```diff
@@ -415,7 +414,7 @@ end = struct
   let save ~argv root =
     let file = filename root in
     Log.info (fun m -> m "Preserving arguments");
-    let args = Array.to_list argv in
+    let args = List.tl (Array.to_list argv) in (* Ignore cmd line*)
```
samoht: The comment is a bit misleading

app/functoria_app.ml
```diff
@@ -484,26 +483,52 @@ end
...
-                build_dir := Some r))
+      build_dir := None;
+      config_file := Fpath.(v "config.ml");
+      ignore (Cmdliner.Term.eval_peek_opts ~argv Cmd.setup_log);
+      ignore (Cmdliner.Term.eval_peek_opts ~argv @@
+              Cmd.config_file (fun c -> config_file := c));
+      ignore (Cmdliner.Term.eval_peek_opts ~argv @@
+              Cmd.build_dir (fun r ->
+                  let (_:bool) = R.get_ok @@ Bos.OS.Dir.create ~path:true r in
+                  build_dir := Some r))
```
samoht: can you remove the extra spaces here?

app/functoria_app.ml
```diff
@@ -607,32 +651,136 @@ module Make (P: S) = struct
...
+  let init_global_state argv =
+      build_dir := None;
+      config_file := Fpath.(v "config.ml");
+      ignore (Cmdliner.Term.eval_peek_opts ~argv Cmd.setup_log);
+      ignore (Cmdliner.Term.eval_peek_opts ~argv @@
+              Cmd.config_file (fun c -> config_file := c));
+      ignore (Cmdliner.Term.eval_peek_opts ~argv @@
+              Cmd.build_dir (fun r ->
+                  let (_:bool) = R.get_ok @@ Bos.OS.Dir.create ~path:true r in
+                  build_dir := Some r))
```
samoht: is it possible to not repeat that piece of code?

app/functoria_app.mli
```diff
@@ -88,8 +88,22 @@ module type S = sig
...
+      to run the configuration builder: command-line arguments will be
+      parsed, and some code will be generated and compiled. *)
+
+  val run_with_argv:
+    ?help_ppf:Format.formatter -> ?err_ppf:Format.formatter ->
+    string array -> unit
+  (** [run_with_argv a] is the same as {!run} but parses [a] instead
+      of the process command line arguments. It also allows to set
+      the error and help channels using [help_ppf] and [err_ppf]. *)
+end
```
samoht: Is it really useful to create a new functor here? Do you see a use-case where users would just want to call `ConfigureBuilder` and where this would make sense to have it separate?

TheLortex: The idea was to clearly separate the two stages of the build. So it's preferable to keep everything under the same module `Make` in order not to break API ?

samoht: Yea I think it's preferable. I don't see how the new API is better for an user point of view (as we just have one user right now who is using both functors anyway)

TheLortex: @samoht I updated the code accordingly

app/functoria_app.ml
```diff
@@ -519,23 +518,239 @@ module Make (P: S) = struct
...
+        let cmd =
+          Bos.Cmd.(v "dune" % "build" % "--no-print-directory" % "--root" % "." % target_file)
+        in
+        write_dune_project_file >>= fun () -> write_dune_file >>= fun () ->
+        Bos.OS.Cmd.run_out cmd |> Bos.OS.Cmd.out_string >>= fun (out, status) ->
+        match snd status with
+        | `Exited 0 ->  Ok (Fpath.(to_string (target_dir / file)))
+        | `Exited _
+        | `Signaled _ ->
+          Format.fprintf Format.str_formatter "error while executing %a\n%s"
```
samoht: This is:

```ocaml
let err = Fmt.strf "error while executing %a\n%s" Bos.Cmd.pp cmd out in
Error (`Invalid_config_ml  x)
```

It would even be better to move the creation of error values in separate functions (so it's easier to spot the different failure modes in one go) but this is not necessary to do for the given PR.
and `| `Signaled _ -> err_invalid_config_ml cmd out`

app/functoria_app.ml
```diff
@@ -519,23 +518,239 @@ module Make (P: S) = struct
...
+    and file_ml = Fpath.basename file_ml_path
+    in
+    Log.info (fun m -> m "Compiling: %a" Fpath.pp file_ml_path);
+    Log.info (fun m -> m "Build dir: %a" Fpath.pp build_dir);
+    let config_dir_to_build_dir =
+      match Fpath.relativize ~root:build_dir config_dir with
+      | Some x -> x
+      | None -> assert false
+    in
+    Bos.OS.Path.matches Fpath.(build_dir / "_build" // cfg + "$(ext)")
```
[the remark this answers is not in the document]
TheLortex: I was legacy so I removed it

app/functoria_app.ml
```diff
@@ -519,23 +518,239 @@ module Make (P: S) = struct
...
-  (* This is set by {!register} *)
-  let configuration = ref None
+  let get_config_dir () = Fpath.(get_cwd () // parent !config_file)
 
-  let default_init = [keys sys_argv]
+  (* STAGE 1 *)
 
-  let register ?packages ?keys ?(init=default_init) name jobs =
+  (* Compile the configuration file. *)
+  let compile file_ml_path =
```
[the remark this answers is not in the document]
TheLortex: I have refactored this function, separated code and added some documentation ! 
Regarding `dune.config` it's probably good to document it in the readme (that needs an update anyway)

app/functoria_app.ml
```diff
@@ -519,23 +518,239 @@ module Make (P: S) = struct
...
-    match !configuration with
-    | None   ->
-      Log.err (fun m -> m "No configuration was registered.") ;
-      Error (`Msg "no configuration file was registered")
-    | Some t -> Ok t
+  let build file =
+    ( match Bos.OS.File.must_exist file with
+      | Ok _ -> Ok ()
+      | Error _ ->
+        Error (`Msg ("configuration file " ^ Fpath.to_string file^ " missing\n"))
```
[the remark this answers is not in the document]
TheLortex: Thank you!

tests/lib/test_app.ml
```diff
@@ -30,7 +30,7 @@ let run cmd =
       failwith err
 
 let rec root path =
-  Bos.OS.File.exists Fpath.(path / "functoria.opam") >>= function
+  Bos.OS.File.exists Fpath.(path / "functoria-runtime.opam") >>= function
```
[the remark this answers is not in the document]
TheLortex: Actually the change is not needed so I reverted it

app/functoria_app.ml
TheLortex: I haven't separated functions but I updated it according to your review

samoht:
Looks good, merging!

PR #172, opened 2019-07-08
Error exit code when `config` subprocess fails

TheLortex:
For now the error is ignored, but it's better to `exit 1` on failure.


samoht:
LGTM

Issue #175, opened 2019-08-03
regressions in functoria 3.0


hannesm:
I finally found the time to try functoria 3.0.0 out, and noticed some regressions //cc @samoht @TheLortex 

- (1) `mirage configure -f config2.ml` used to work, now I get:
```
File "config/dune", line 4, characters 11-17:
4 |   (modules config)
               ^^^^^^
Error: Module Config doesn't exist.
error while executing dune build --no-print-directory --root
                        /usr/home/hannes/devel/mirage/mirage-skeleton/device-usage/network/./
                        @config
```

- (2) a typo/error in `config.ml` leads to an error location in `config/config.ml` -- that's unexpected since I never did anything in the `config` subdirectory (previously the error location was in `config.ml`)
```
File "config/config.ml", line 11, characters 0-3:
11 | foo
     ^^^
Error: The function applied to this argument has type [...]
```

- (3) a unikernel (such as https://git.robur.io/?p=openvpn.git;a=tree;f=mirage-client;h=c69bca6813e17f462ed9373806113afc9a8ccaf5;hb=HEAD) that uses the subdirectory `config` to `crunch` the configuration:
  - (a) removed my configuration and .gitignore during `mirage clean` (that's unexpected and a bit too eager, how will I now get my configuration back?)
  - (b) after mirage configure && mirage build the crunched `static1.ml` contains more files than expected
~> maybe functoria should be a bit more careful to only remove stuff it generated itself? (and/or warn/refuse to do anything if a hand-crafted `config` directory exists?; or reside in some `_YYY` directory (as already used by mirage: `_build` / `_build-ukvm` / `_build-solo5-hvt`))

- (4) app-info is broken: a `mirage configure && mirage build` in mirage-skeleton/tutorial/app_info leads to:
```
Fatal error: Invalid character in package name "\027[01;04mlwt\027[0m"

run ['opam' 'config' 'subst' 'info_gen.ml']: exited with 99run ['dune' 'exec' '--root'
     '/usr/home/hannes/devel/mirage/mirage-skeleton/tutorial/app_info/./'
     '--' 'config/config.exe' 'build']: exited with 1
```

- (5) `mirage configure -b tst && mirage build -b tst` leads to an error:
```
Error: Path . has already been scanned. Cannot scan it again through symlink
tst
error while executing dune build --no-print-directory --root
                        /usr/home/hannes/devel/mirage/mirage-skeleton/tutorial/app_info/tst
                        @config
```

[...]

PR #177, opened 2019-08-12
Invoke "opam list" with --color=never

ehmry:
This is a fix for a problem with color and undeline escape sequences in the output of `opam list`.


hannesm:
this looks good to me and addresses #175 point (4).

hannesm:
only thing I'm not sure about is whether there is a nicer way -- could the parent process which executes the `opam list` do this without a real terminal (i.e. as before), so `opam list` will automatically detect to not emit ansi colour sequences.

samoht:
> only thing I'm not sure about is whether there is a nicer way
I think controlling `opam` with an explicit color argument is the right way to go as we won't have to guess how opam detects pseudo-terminals (and it changed between opam1 and opam2 if I remember correctly)

PR #180, opened 2019-10-14
Command exit master

hannesm:
that's #179 but targeting master instead of 2.2!


hannesm:
merging to stay consistent with released 2.2 series

PR #181, opened 2019-10-22
Filename.extension is available on 4.04

samoht:


hannesm:
we can as well bump to 4.06.0 directly and be in the safe-string land by default! :)

hannesm:
according to https://ci.ocaml.org/log/saved/docker-run-0df0bbb43658b1462dbb2fb3edece65c/257e3c0e9480764c4d68256b58f6316d5011f63f functoria-runtime needs to be adjusted as well.

PR #183, opened 2019-11-03
remove custom opam version comparison which is inconsistent with opam, fix #149

### LINKED ISSUE #149: version comparison is inconsistent with opam
while opam's version comparison is inherited from debian, and licensed under LGPL 2.1 (while functoria is ISC licensed), the adhoc version comparison in functoria is incomplete and inconsistent:
- in functoria, a version may only be numbers and `.`
- in opam, `1.0 != 1.0.0`, while in functoria `1.0 = 1.0.0`

this may be an issue in the future, and since I wrote the code last year I wanted to inform others that there may be issues at the end of the road.  I guess the solution [tm] is to develop a version comparison library _from scratch_ with a permissive license, and use this in functoria (and in the best case also in opam).  another solution is to use the opam version comparison (and depend on opam, which would make this library LGPL IIUC).

this is not really urgent - as long as we stick to versions which only contain numbers and `.` (and do not release both `foo.1.0` and `foo.1.0.0`) we're safe.

hannesm:
instead of merging min and max constraints, collect them in a set and output
 all of them, leaving the merge to opam

i.e. consider the collected bounds:
- package ~min:"1.0" ~max:"2.0" "a"
- package ~min:"1.2" ~max:"3.0" "a"
- package ~min:"1.5" ~max:"2.0" "a"

it used to merge them, and produce
 "a" {>= "1.5" & < "2.0"}

now, it'll output
 "a" {>= "1.0" & >= "1.2" & >= "1.5" & < "2.0" & < "3.0"}

the previous approach failed on various version identifiers that are valid in
opam, such as "v1.0", "1.2.3-99", "2.0~beta" -- since the functoria version
parser only allowed dot "." and numbers.

### DIFF
--- a/lib/functoria.ml
+++ b/lib/functoria.ml
@@ -26,67 +26,28 @@ type package = {
   pin : string option ;
   build : bool ;
   ocamlfind : String.Set.t ;
-  min : string option ;
-  max : string option
+  min : String.Set.t ;
+  max : String.Set.t ;
 }
 
 module Package = struct
-  (* we could have copied code from opam, but that's LGPL *)
-  let version_of_string s =
-    let r = List.fold_left (fun acc v ->
-        match Astring.String.to_int v with
-        | Some n -> n :: acc
-        | None -> invalid_arg "cannot parse version number")
-        []
-        (Astring.String.cuts ~sep:"." s)
-    in
-    List.rev r
-
-  let compare_version v1 v2 =
-    let rec cmp a b =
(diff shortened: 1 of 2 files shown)


hannesm:
a successful mirage-skeleton travis run using this branch with mirage emitting a constraint `>= "0.5.4-2"`is at https://travis-ci.org/mirage/mirage/builds/606727342

PR #184, opened 2019-11-03
opam: do not use make, but dune runtest and dune exec directly

hannesm:
suggested by @kit-ty-kate in ocaml/opam-repository#15099


hannesm:
this is mainly adapted from opam-repository, and CI is happy (and still executes the tests!) merging

PR #187, opened 2019-11-26
dune exec only a single job

hannesm:
The observed issue (in CI https://travis-ci.org/mirage/mirage-nat/jobs/607896490?utm_medium=notification&utm_source=github_status and locally)
```
File "unikernel.ml", line 74, characters 27-47:
Error: Unbound value Key_gen.private_ipv4
```

Another observation is in [this PR](https://github.com/roburio/openvpn/pull/6) where CI ran twice (push and pr) with diverging output:
- failure (qubes build) https://travis-ci.org/roburio/openvpn/builds/616130870?utm_source=github_status&utm_medium=notification
- success https://travis-ci.org/roburio/openvpn/builds/616130861?utm_source=github_status&utm_medium=notification

Now, this is the same code in the CI runs (and same installed libraries), thus there must be a race condition somewhere (in this examples, it is calling `create_ipv4` twice with different `~group` parameters). I managed to reproduce this locally, sometimes with the Unix target, other times with the qubes target).

After reading the `_build/log`, the configuration phase inferred two jobs. With the change in this PR, I cannot reproduce the race condition (and thus would like this to be merged -- I'm happy/open if someone else investigates where/when concurrency is an issue for functoria, but it won't be me). FWIW, cloning the mirage-nat repository, using the `easy` branch from hannesm/mirage-nat, and running `mirage clean ; mirage configure -vv && grep ipv4\  key_gen.ml` in a tight loop, erroring if grep only finds two lines (i.e. the private_ipv4 is missing), a good outcome is grep finding 4 lines.

thanks for reading so far. /cc @mirage/core @TheLortex @samoht

PR #188, opened 2019-12-04
Fix equality for 'a impl values

samoht:
This fixes random issues with `mirage configure` when multiple keys
sharing the same name. Replaces #187

### DIFF
--- a/lib/functoria_key.ml
+++ b/lib/functoria_key.ml
@@ -211,6 +211,7 @@ end
 
 type t = Set.elt = Any: 'a key -> t
 let compare = Set.M.compare
+let equal (Any x) (Any y) = String.equal x.name y.name
 
 module Alias = struct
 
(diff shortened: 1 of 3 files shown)


lib/functoria_key.ml
```diff
@@ -211,6 +211,7 @@ end
 
 type t = Set.elt = Any: 'a key -> t
 let compare = Set.M.compare
+let equal (Any x) (Any y) = x.name = y.name
```
Drup: `String.equal` ?

Drup: Or, alternatively, reuse the `compare` defined earlier.

Drup:
I'm pretty sure that was initially on purpose so that we could share the module implementation of configurable, even if they are instanciated twice with different key names, and it seems that was not a good idea at all. Well spotted, feel free to merge. :)

Issue #182, opened 2019-11-01
regression in functoria 3.0


hannesm:
from ocaml-tls in mirage/examples:
```
] mirage configure
File "config.ml", lines 7-9, characters 4-25:
7 | ....match Sys.getenv "BUILD" with
8 |     | "client" -> `Client
9 |     | "server" -> `Server
Error (warning 8): this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
""
run ['dune' 'exec' '--root'
     '/usr/home/hannes/devel/mirage/ocaml-tls/mirage/example' '--'
     './config.exe' 'configure']: exited with 1
```

--> mirage configure used to work, now it fails due to an incomplete pattern match. I'm not sure whether this should be expected.

hannesm:
I'm fine with `warnings are errors` (and we can close this if everyone agrees)

samoht:
The `warning as errors` are enabled in dune's dev mode only. Currently there is no way to use the release mode but we should probably add something -- for production unikernel vs. dev setups.

hannesm:
this would be great to have fixed and released (with a mirage3-compatible functoria) rather sooner than later, because only then we are able to deprecate bindings for config.ml (and remove them with 4.0). thanks!

hannesm:
thanks for fixing this in #193 / functoria 3.1.0 @samoht

PR #193, opened 2020-03-09
Do not compile config.ml with -warn-error by default

samoht:
A tentative fix for https://github.com/mirage/functoria/issues/182

Both functoria and mirage are defining the same flag, so I'm not sure if this will work very well without a patch to mirage too (or we will have to rename that flag).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+## v3.1.0 (2020-03-09)
+
+* Always use `-warn-error -A` when compiling `config.ml`. This allows
+  to use deprecated devices without failing (#193, @samoht)
+
 ## v3.0.3 (2019-12-17)
 
 * Fix equality for `'a impl` values, which caused issue in `mirage configure`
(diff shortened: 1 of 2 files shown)


hannesm:
thanks @samoht -- another solution would be to never `warn_error` (and thus revert the behaviour to functoria pre-3). in functoria-4/mirage-4 (developed on mirage#master), we can find a cleaner solution (enable warn-error by default, as dune does, and have a single command-line flag to disable warn-error). this would allow us to not have to struggle with lower bounds and cut unnecessary releases. what do you think?

samoht:
yes that's a good idea, that'd make that patch even simpler. I'll push this shortly.

samoht:
Updated, and tested on a small example : the deprecation messages in `config.ml` are still printed out (annoyingly they are duplicated) but at least `mirage configure` complete fine.

hannesm:
cool, thanks :)

hannesm:
should we merge + release this then?

samoht:
Yes, I'm making a release.

samoht:
functoria-3.1.0 has been submitted: https://github.com/ocaml/opam-repository/pull/15968

PR #197, opened 2021-02-22
Backport #1210 to the v3 branch

samoht:
I'm not sure it's a good idea to merge this, but it shows what has to be done to backport the graph removal patch to v3.

The API is changing a bit (as it adds a type parameter to the base classes) and I haven't tested the changes at all.


dinosaure:
> The API is changing a bit (as it adds a type parameter to the base classes) and I haven't tested the changes at all.

I don't understand why it's necessary, it's possible to keep the same interface and do a smooth upgrade for MirageOS 3?

dinosaure:
According to mirage/mirage#1214, this PR can be close.

